PR #82, opened 2015-01-30
Add /memory-stats endpoint to the API.

alext:
To help with debugging memory usage.

This could be useful, but It's worth noting that the `ReadMemoryStats()` call has to stop the world[1] in order to gather the stats, which could have a performance impact if it's hit rapidly.

[1] http://golang.org/src/runtime/mem.go?s=2417:2447#L63


mattbostock:
:+1:

PR #86, opened 2015-03-18
Update the mgo package.

alext:
This includes changing to its new import path (See
http://labix.org/mgo for more details).

### DIFF
--- a/router.go
+++ b/router.go
@@ -2,14 +2,15 @@ package main
 
 import (
 	"fmt"
-	"github.com/alphagov/router/handlers"
-	"github.com/alphagov/router/logger"
-	"github.com/alphagov/router/triemux"
-	"labix.org/v2/mgo"
 	"net/http"
 	"net/url"
 	"sync"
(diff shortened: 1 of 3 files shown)


router.go
```diff
@@ -2,14 +2,15 @@ package main
...
-	"github.com/alphagov/router/handlers"
-	"github.com/alphagov/router/logger"
-	"github.com/alphagov/router/triemux"
-	"labix.org/v2/mgo"
 	"net/http"
 	"net/url"
 	"sync"
 	"time"
+
+	"github.com/alphagov/router/handlers"
```
ajlanghorn: Should these not be alphabetised inside the `import`?

alext: They are, but within groups.  There's an emerging convention to put the imports into groups, with all the stdlib imports first, and all the 3rd party imports separately.

Additionally, the [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports) tool automatically arranges them this way.

PR #89, opened 2015-06-04
Return a 503 when running with no routes.

alext:
If the connection to Mongodb fails when the router first starts up, it
will come up with an empty routing table.  Previously this would return
a 404 for all requests.  These would be served to end users and cached
upstream.

Running with an empty routing table is always going to be erroneous, so
this changes it to return a 503's in this situation. This will cause
upstream caches to serve from stale, and additionally won't be cached
anywhere.

### DIFF
--- a/integration_tests/backend_selection_test.go
+++ b/integration_tests/backend_selection_test.go
@@ -10,15 +10,6 @@ import (
 
 var _ = Describe("Backend selection", func() {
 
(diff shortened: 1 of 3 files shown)


mattbostock:
:+1:

alexmuller:
This is great, thanks :+1:

PR #90, opened 2015-06-08
Add version information to help output. 

alext:
This will allow us to determine what version of the router a given
binary is, and additionally, which version of Go was used to compile it.

### DIFF
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -4,3 +4,4 @@ set -eu
 
 make
 make test
+./router -version
--- a/version.go
+++ b/version.go
@@ -1,0 +1,13 @@
+package main
+
+import (
+	"fmt"
+	"runtime"
+)
+
+// populated by -ldflags from Makefile
+var version = "unknown"
+
+func versionInfo() string {
+	return fmt.Sprintf("build: %s (compiler: %s)", version, runtime.Version())
+}
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,18 @@
 .PHONY: build run test clean
 
 BINARY := router
-BUILDFILES := router.go main.go router_api.go
+BUILDFILES := router.go main.go router_api.go version.go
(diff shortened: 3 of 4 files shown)


jenkins.sh
```diff
@@ -2,5 +2,15 @@
 set -x
 set -eu
 
+if [ -n "${GIT_COMMIT-}" ]; then
+  sed -i "s/buildCommit = .*$/buildCommit = \"${GIT_COMMIT}\"/" version.go
```
mattbostock: Could we use `-ldflags` to set the value at compile time rather than using `sed`?

I've done this in one of my personal projects:
https://github.com/mattbostock/leavediary/blob/f27014f9be4dffefeead448c67f7f12778730867/Makefile#L1-L7

alext: Nice, I'll have a look at that.  Thanks.

ajlanghorn: @alext Did you get anywhere on this? I'm on a PR review rampage.

alext: Updated to use this approach.

version.go
```diff
@@ -0,0 +1,13 @@
+package main
+
+import (
+	"fmt"
+	"runtime"
+)
+
+// populated by -ldflags from Makefile
+var version = "unknown"
```
mattbostock: You could have a runtime check to make sure that the version is populated via make:
https://github.com/mattbostock/leavediary/blob/adc0179a76e0fe6281506041baf103bbb373b052/main.go#L86-L88

mattbostock:
One nitpicking suggestion above, but this is ready to merge as-is :+1:

version.go
alext: I thought about that, but it would make it difficult for people using `go run` etc...  I'm not sure this is something that would be a problem in reality.  The builds where this matters are the ones we deploy, and they'll always be built on the CI servers using make.

PR #91, opened 2015-06-08
Rate-limit route reloading.

danielroseman:
Although ReloadRoutes ensures that routes continue to be served while they are being reloaded, the reload itself still takes some time, and the client making the request will be blocked until the reload is complete. This means for example that content-store can become unresponsive when it tries to register a large number of routes. 

This change implements a rate limit, set by the ROUTER_RELOAD_INTERVAL env var (default 1s), so that a reload request waits for that interval before doing the actual reload, and any further requests within that time do not trigger an extra reload.

Local testing shows that this speeds up the general time to register 900 content items by about 15%, but it's not possible in that environment to test how it affects the responsiveness of content-store in this scenario - that may need to be done in preview.

### DIFF
--- a/integration_tests/reload_api_test.go
+++ b/integration_tests/reload_api_test.go
@@ -3,6 +3,7 @@ package integration
 import (
 	"crypto/sha1"
 	"fmt"
+	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -30,6 +31,34 @@ var _ = Describe("reload API endpoint", func() {
 			resp := doRequest(newRequest("GET", routerAPIURL("/reload")))
 			Expect(resp.StatusCode).To(Equal(405))
 			Expect(resp.Header.Get("Allow")).To(Equal("POST"))
+		})
+
+		Context("with a non-zero reload interval", func() {
+			newRouterPort := 7999
+			newApiPort := 8000
+			BeforeEach(func() {
+				err := startRouter(newRouterPort, newApiPort, map[string]string{"ROUTER_RELOAD_INTERVAL": "1s"})
+				Expect(err).To(BeNil())
+			})
+
+			AfterEach(func() {
+				stopRouter(newRouterPort)
+			})
+
+			It("should return 'already in progress' for requests within timeout", func() {
+				resp := doRequest(newRequest("POST", routerURL("/reload", newApiPort)))
+				resp2 := doRequest(newRequest("POST", routerURL("/reload", newApiPort)))
+				Expect(readBody(resp)).To(Equal("Reload triggered"))
+				Expect(readBody(resp2)).To(Equal("Reload already in progress"))
+			})
+
+			It("should return 'triggered' for requests after timeout", func() {
+				resp := doRequest(newRequest("POST", routerURL("/reload", newApiPort)))
+				time.Sleep(time.Second * 2)
+				resp2 := doRequest(newRequest("POST", routerURL("/reload", newApiPort)))
+				Expect(readBody(resp)).To(Equal("Reload triggered"))
+				Expect(readBody(resp2)).To(Equal("Reload triggered"))
+			})
 		})
 	})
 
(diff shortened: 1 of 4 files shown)


integration_tests/reload_api_test.go
```diff
@@ -31,6 +32,39 @@ var _ = Describe("reload API endpoint", func() {
...
+			BeforeEach(func() {
+				err := startRouter(newRouterPort, newApiPort, map[string]string{"ROUTER_RELOAD_INTERVAL": "1s"})
+				Expect(err).To(BeNil())
+			})
+
+			AfterEach(func() {
+				stopRouter(newRouterPort)
+				// stopRouter doesn't actually seem to stop the router process for
+				// some reason. The only thing to do is to bump the port numbers and
+				// start yet another instance for the next test.
```
alext: It would be good to understand why this is an issue here.  It works correctly in [another test](https://github.com/alphagov/router/blob/master/integration_tests/proxy_function_test.go#L87-L102).  I'm worried that this could leave orphaned processes around on the CI machines, which will then interfere with future test runs etc.

integration_tests/router_support.go
```diff
@@ -32,6 +32,7 @@ func reloadRoutes(optionalPort ...int) {
 	resp, err := http.Post(fmt.Sprintf("http://localhost:%d/reload", port), "", nil)
 	Expect(err).To(BeNil())
 	Expect(resp.StatusCode).To(Equal(200))
+	time.Sleep(time.Millisecond * 50)
```
alext: Would be good to have a comment explaining the need for this sleep.

router_api.go
```diff
@@ -4,9 +4,26 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+	reloadChan := make(chan bool)
+	go func() {
+		// Rate-limit reloads to 1 per RELOAD_INTERVAL.
+		// This goroutine blocks until it receives a message on reloadChan, then
+		// waits for the timeout before calling reload.
+		for {
+			<-reloadChan
```
alext: This would be more idiomatic as:

```go
for range reloadChan {
    //...
}
```
See range clauses in the [for statement documentation](http://golang.org/ref/spec#For_statements) for more details.

integration_tests/router_support.go
KushalP: Why is this 50ms sleep required?

alext: because route reloading is no longer synchronous.

alext:
I'm a little unsure about whether this is the correct thing to be doing.  It does change the behaviour of the reload endpoint so that it's no longer synchronous (at the least we should be returning a 202 status code now).  At the moment, the router (and the router-api) have the nice property that when the reload call has returned, you know that the new routes are active.

As an aside, I'm not sure that the delay is necessary here, it's simply the process of making it asynchronous that's helped the downstream apps.  The router itself will happily continue to serve requests [while it's reload endpoint is being hammered](https://github.com/alphagov/router/blob/master/integration_tests/performance_test.go#L41-L58).

It seems that the root of the problem is the fact the both write and read requests to the content-store share a pool of workers.  That feels like a better place to approach solving this problem.  It would be possible to implement a rate-limit for write requests in the publishing-api.  That could ensure that at most a given number of write requests were active to the content-store, and therefore ensure that there were always some workers available to serve read requests.

boffbowsh:
We discussed this problem recently in relation to the performance of republishing thousands of documents from Whitehall. The speed improvement this will offer is essential in order to complete these republishing tasks in a timely manner. 

Alex's comment against this change doesn't seem to be relevant now that publishing-api is no longer a dumb proxy for the content stores and is an app in its own right.

This has no user-facing performance impact, rather it removes what is essentially an artificial bottleneck on republishing.

I've rebased this on master and force pushed, and I've added a commit to make a POST to `/reload` return a 202 as Alex suggested.

boffbowsh:
Reopened as #104

PR #92, opened 2015-07-21
Reduce redirect cache TTL to 30 mins.

alext:
The current cache TTL of 1 day can cause problems when redirects are
changed, or even replaced with a regular route.

Reducing this to 30 minutes will make it less surprising (it's more
consistent with the rest of the stack). It shouldn't make much
difference to the load on the servers because redirects are very
cheap to serve.


mattbostock:
:+1:

PR #97, opened 2015-09-04
Fix dependency installation.

alext:
The build is currently failing on master because it can't find quantile
(a dependency of vegeta). I'm not sure why this was working before, and
has stopped now, but we should be pinning this anyway.

This should fix build failures like https://ci-new.alphagov.co.uk/job/govuk_router/114/console


mattbostock:
It could be vegeta only recently started depending on that library.

alext:
> It could be vegeta only recently started depending on that library.

We're pinning to a specific tag for vegeta though, so that shouldn't be the case.

PR #98, opened 2015-09-07
Support special characters (eg space) in paths.

### LINKED ISSUE #96: Bug: incoming URLs are decoded before being looked up
Problem example: `https://www.gov.uk/vehicle%20tax` should match against the route [defined here](https://github.gds/gds/router-data/blob/92eac5877c47411005957211bf12ee51b0a503c3/data/typos.csv#L28) but doesn't.

If a request URL has a space, encoded as `%20`, it won't be matched a route. This is because the router is decoding the request URL before matching against the encoded list [`here`](https://github.com/alphagov/router/blob/e678977823c58e40a24a1ee53ac0cf315227797e/triemux/mux.go#L44).

Routes in the database should be in their encoded form.

One option is to switch the call to something which returns the encoded version, but this will probably require an upgrade to Go 1.5. @alext has started doing some thinking about this.

alext:
The database contains these special routes in their %-encoded form, but
the paths we were comparing against were in their decoded form, and
therefore didn't match. This updates our matching code to decode the
paths from the database before comparing against the request URL so that
we're comparing equivalent forms.

This is needed to support some redirects for common typos (eg `/vehicle tax`).

/cc @jamiecobbett 

Fixes #96

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,9 @@ The `routes` collection uses the following data structure:
 }
 ```
 
+Incoming paths with special characters must be in their % encoded form in the
+database (eg spaces must be stored as `%20`).
+
 The behaviour is determined by `handler`. See below for extra fields
 corresponding to `handler` types.
 
--- a/integration_tests/redirect_test.go
+++ b/integration_tests/redirect_test.go
@@ -103,6 +103,15 @@ var _ = Describe("Redirection", func() {
 				time.Second,
 			))
 		})
+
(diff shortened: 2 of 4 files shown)


jamiecobbett:
LGTM, but best if someone who has used Go more than I do takes a look.

mattbostock:
:+1:

fofr:
Very happy to see this :smile: – it will fix the most frequent 404 error, about 300 per day.

PR #102, opened 2015-10-19
Add new behaviour for prefix redirects

boffbowsh:
https://trello.com/c/BYS3dxxV/118-panopticon-unpublishings-should-redirect-without-keeping-subpath-medium

By default, prefix redirects preserve the path sections after the
prefix and add them to the redirect destination. This isn’t always the
desired behaviour.
This adds a new `prefix_mode` field for redirect routes, set to
`preserve` (the existing behaviour) or `ignore`. If the field isn’t
present in the route document, this will default to `preserve` for
prefix routes.
If the mode is set to `ignore`, the path sections after the prefix will
be discarded and all requests will be redirected to the exact target.

Includes test refactoring to support this work. See individual commits for details.

/cc @alext @bradwright

### DIFF
--- a/handlers/redirect_handler.go
+++ b/handlers/redirect_handler.go
@@ -14,7 +14,7 @@ func NewRedirectHandler(source, target string, prefix bool, preserve bool, tempo
 	if temporary {
 		statusMoved = http.StatusFound
 	}
-	if prefix && preserve {
+	if preserve {
 		return &pathPreservingRedirectHandler{source, target, statusMoved}
 	}
 	return &redirectHandler{target, statusMoved}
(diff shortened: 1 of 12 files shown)


alext:
Code changes look good.

I do wonder whether it would be better to generalise this a little more so that it's not specific to prefix redirects. At the moment, we have a couple of cases where we want to do an exact redirect that preserves query params etc. That's currently not possible, and we're having to work around this limitation by using prefix redirects.

If we separated the redirect behaviour from the route type completely, then this would become more flexible. We'd need to add some logic to router-api so that if the preserving behaviour wasn't specified, its default value would be set so as to maintain the existing behaviour.

boffbowsh:
Done, @alext

PR #104, opened 2016-02-26
Rate-limit route reloading

boffbowsh:
Originally opened as #91 by @danielroseman 

> Although ReloadRoutes ensures that routes continue to be served while they are being reloaded, the reload itself still takes some time, and the client making the request will be blocked until the reload is complete. This means for example that content-store can become unresponsive when it tries to register a large number of routes.

> This change implements a rate limit, set by the ROUTER_RELOAD_INTERVAL env var (default 1s), so that a reload request waits for that interval before doing the actual reload, and any further requests within that time do not trigger an extra reload.

> Local testing shows that this speeds up the general time to register 900 content items by about 15%, but it's not possible in that environment to test how it affects the responsiveness of content-store in this scenario - that may need to be done in preview.

@alext commented previously:

> I'm a little unsure about whether this is the correct thing to be doing. It does change the behaviour of the reload endpoint so that it's no longer synchronous (at the least we should be returning a 202 status code now). At the moment, the router (and the router-api) have the nice property that when the reload call has returned, you know that the new routes are active.

> As an aside, I'm not sure that the delay is necessary here, it's simply the process of making it asynchronous that's helped the downstream apps. The router itself will happily continue to serve requests while it's reload endpoint is being hammered.

> It seems that the root of the problem is the fact the both write and read requests to the content-store share a pool of workers. That feels like a better place to approach solving this problem. It would be possible to implement a rate-limit for write requests in the publishing-api. That could ensure that at most a given number of write requests were active to the content-store, and therefore ensure that there were always some workers available to serve read requests.

We discussed this problem recently in relation to the performance of republishing thousands of documents from Whitehall. The speed improvement this will offer is essential in order to complete these republishing tasks in a timely manner.

Alex's comment against this change doesn't seem to be relevant now that publishing-api is no longer a dumb proxy for the content stores and is an app in its own right.

This has no user-facing performance impact, rather it removes what is essentially an artificial bottleneck on republishing.

I've rebased this on master and force pushed, and I've added a commit to make a POST to /reload return a 202 as Alex suggested.

### DIFF
--- a/integration_tests/reload_api_test.go
+++ b/integration_tests/reload_api_test.go
@@ -3,6 +3,7 @@ package integration
 import (
 	"crypto/sha1"
 	"fmt"
+	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -30,6 +31,34 @@ var _ = Describe("reload API endpoint", func() {
 			resp := doRequest(newRequest("GET", routerAPIURL("/reload")))
 			Expect(resp.StatusCode).To(Equal(405))
 			Expect(resp.Header.Get("Allow")).To(Equal("POST"))
+		})
+
+		Context("with a non-zero reload interval", func() {
+			newRouterPort := 7999
+			newApiPort := 8000
+			BeforeEach(func() {
+				err := startRouter(newRouterPort, newApiPort, map[string]string{"ROUTER_RELOAD_INTERVAL": "100ms"})
+				Expect(err).To(BeNil())
+			})
+
+			AfterEach(func() {
+				stopRouter(newRouterPort)
+			})
+
+			It("should return 'already in progress' for requests within timeout", func() {
+				resp := doRequest(newRequest("POST", routerURL("/reload", newApiPort)))
+				resp2 := doRequest(newRequest("POST", routerURL("/reload", newApiPort)))
+				Expect(readBody(resp)).To(Equal("Reload triggered"))
+				Expect(readBody(resp2)).To(Equal("Reload already in progress"))
+			})
+
+			It("should return 'triggered' for requests after timeout", func() {
+				resp := doRequest(newRequest("POST", routerURL("/reload", newApiPort)))
+				time.Sleep(time.Second * 2)
+				resp2 := doRequest(newRequest("POST", routerURL("/reload", newApiPort)))
+				Expect(readBody(resp)).To(Equal("Reload triggered"))
+				Expect(readBody(resp2)).To(Equal("Reload triggered"))
+			})
 		})
 	})
 
--- a/integration_tests/reload_api_test.go
+++ b/integration_tests/reload_api_test.go
@@ -12,9 +12,9 @@ import (
 var _ = Describe("reload API endpoint", func() {
 
 	Describe("request handling", func() {
-		It("should return 200 for POST /reload", func() {
+		It("should return 202 for POST /reload", func() {
 			resp := doRequest(newRequest("POST", routerAPIURL("/reload")))
-			Expect(resp.StatusCode).To(Equal(200))
+			Expect(resp.StatusCode).To(Equal(202))
 		})
 
 		It("should return 404 for POST /foo", func() {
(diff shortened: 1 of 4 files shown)


router_api.go
```diff
@@ -15,8 +31,17 @@ func newAPIHandler(rout *Router) http.Handler {
...
+		// before reloading, or do nothing if one is already in progress.
+		select {
+		case reloadChan <- true:
+			logInfo("router: reload triggered")
+			w.WriteHeader(http.StatusAccepted)
+			w.Write([]byte("Reload triggered"))
+		default:
+			logInfo("router: reload already in progress")
+			w.Write([]byte("Reload already in progress"))
+		}
```
alext: This looks a little racy to me. I think there's a case here, where a reload request could come in after the reload has completed, but not actually trigger a reload.

I think we need some sort of sentinel that's only in place around the sleep, but reset before the actual reload starts so that any requests that some in after the reload has started will trigger another sleep and reload once it's complete.

main.go
```diff
@@ -114,7 +116,10 @@ func main() {
 	go catchListenAndServe(pubAddr, rout, "proxy", wg)
 	logInfo("router: listening for requests on " + pubAddr)
 
-	api := newAPIHandler(rout)
+	api, err := newAPIHandler(rout)
```
alext: I think it would be cleaner to pass in `reloadInterval` here.  This will also make it consistent with the `NewRouter` call above.

integration_tests/reload_api_test.go
```diff
@@ -31,6 +32,34 @@ var _ = Describe("reload API endpoint", func() {
 			Expect(resp.StatusCode).To(Equal(405))
 			Expect(resp.Header.Get("Allow")).To(Equal("POST"))
 		})
+
+		Context("with a non-zero reload interval", func() {
+			newRouterPort := 7999
+			newApiPort := 8000
+			BeforeEach(func() {
+				err := startRouter(newRouterPort, newApiPort, map[string]string{"ROUTER_RELOAD_INTERVAL": "1s"})
```
alext: I think a second here is overkill for the tests and will only slow them down unnecessarily.  It should be sufficient to sue something like 100ms or possibly even less.

router_api.go
rboulton: I'm not sure there's a race here: before the sleep starts the value should have been read from the channel, so the channel will be already be empty and a subsequent call to send a value to the channel will work.  The channel can't then be cleared again until that new value is read.

I think this has be true, because every value read from the channel triggers a reload, and values are removed from the channel atomically compared to checking whether the channel is empty. (Though this atomicity isn't actually required, you just need to be sure that the values are removed from the channel before the reload start.)

integration_tests/reload_api_test.go
boffbowsh: Updated.

main.go
boffbowsh: Added in 03c20fe499f0cce995f66fcd93c0003794b7089f, is that what you meant?

router_api.go
boffbowsh: With my limited go-fu, @rboulton's argument makes sense to me. Once the value is read in the for loop, there's a second plus reload time before the next loop. But as the value has been read and cleared at the for condition, it's immediately available to be written to again by an incoming request.

alext: The problem here is that it's using an unbuffered channel, so a send on the channel will only succeed when there's something waiting to receive - https://golang.org/ref/spec#Send_statements.

boffbowsh: Could we make this a channel with a capacity of 1?

On first request, the loop executes immediately and takes about 3 seconds.
In those three seconds, another request is received, a message is sent to the channel and is "queued" for the next loop execution.
Another request is received in those 3 seconds. The channel can't receive this message so the call fails, but there's still a message in the channel.
When the loop finishes, it executes immediately again due to the message in the channel, and clears the channel for future requests.

alext: That would address the race condition, but would reduce the efficiency, because 2 requests in quick succession will always trigger 2 reloads. I reckon there's a way we can get the lock period to only apply around the sleep period, but it might involve a bit more code.  Let's have a chat this afternoon to explore this.

boffbowsh:
Closing in favour of #105

PR #105, opened 2016-03-02
Make router reloading asynchronous

boffbowsh:
Supersedes #104 

Use a channel based mechanism for actioning router reload requests from
router-api. POST requests to /reload will now result in a message being
sent to a channel which will cause a goroutine to reload the routes.

The channel has a capacity of 1 in order to handle a request arriving
while routes are being reloaded. In this situation, the message will be
“queued” in the channel ready to be read again once the reload for loop
is complete.

The response for /reload is changed to be a 202 to reflect that the
reload is no longer synchronous.

Opinions are welcome on the test case.

- [ ] Deploy https://github.com/alphagov/router-api/pull/81 first

### DIFF
--- a/integration_tests/reload_api_test.go
+++ b/integration_tests/reload_api_test.go
@@ -11,9 +11,10 @@ import (
 var _ = Describe("reload API endpoint", func() {
 
 	Describe("request handling", func() {
-		It("should return 200 for POST /reload", func() {
+		It("should return 202 for POST /reload", func() {
 			resp := doRequest(newRequest("POST", routerAPIURL("/reload")))
-			Expect(resp.StatusCode).To(Equal(200))
+			Expect(resp.StatusCode).To(Equal(202))
+			Expect(readBody(resp)).To(Equal("Reload queued"))
 		})
 
 		It("should return 404 for POST /foo", func() {
--- a/integration_tests/reload_api_test.go
+++ b/integration_tests/reload_api_test.go
@@ -3,6 +3,7 @@ package integration
 import (
 	"crypto/sha1"
 	"fmt"
+	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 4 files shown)


boffbowsh:
@alext / @danielroseman

integration_tests/reload_api_test.go
```diff
@@ -31,6 +33,28 @@ var _ = Describe("reload API endpoint", func() {
...
+			doRequest(newRequest("POST", routerAPIURL("/reload")))
+			end := time.Now()
+			duration := end.Sub(start)
+
+			Expect(duration.Nanoseconds()).To(BeNumerically("<", 1000000))
+
+			addRoute("/bar", NewRedirectRoute("/qux", "prefix"))
+			doRequest(newRequest("POST", routerAPIURL("/reload")))
+
+			time.Sleep(time.Second * 1)
```
alext: Minor detail: does this need to sleep for a full second?  If not, it worth shortening it to avoid making the tests slower than they need to be.

alext:
This looks good to me. One small comment about the tests, but that probably shouldn't block this.

I'll leave it to @danielroseman to merge when he's happy.

integration_tests/reload_api_test.go
boffbowsh: Fixed and rebased

benlovell:
:ship:

router_api.go
```diff
@@ -15,8 +29,15 @@ func newAPIHandler(rout *Router) http.Handler {
 			w.WriteHeader(http.StatusMethodNotAllowed)
 			return
 		}
-
-		rout.ReloadRoutes()
+		// Send a message to the reload goroutine, which will start a new timeout
```
mattbostock: I think the comment about the timeout is now out of date?

boffbowsh: Well spotted. Addressed in https://github.com/alphagov/router/pull/106

PR #107, opened 2016-07-18
changed install command for gom

Seraphiana:
- go install command doesn't work for gom, seems like you need to
use get instead


surminus:
:+1:

PR #108, opened 2016-07-19
Upgrade to Go 1.6.3

timblair:
Version bump due to CVE-2016-5386 (or ["httpoxy"](https://httpoxy.org/)).

As Go 1.7 is just about to be released, the 1.5.x branch is no longer receiving security patches as per the [Go release cycle](https://github.com/golang/go/wiki/Go-Release-Cycle#release-maintenance), so here we're upgrading to 1.6.3.

Also moves dependency management away from Gom and over to Godep, and updates all dependencies to their latest versions.

/cc @gpeng


Makefile
```diff
@@ -18,23 +14,18 @@ endif
 
 build: $(BINARY)
 
-run: vendor/stamp
-	gom run $(BUILDFILES)
+run:
+	godep go run $(BUILDFILES)
```
alext: It shouldn't be necessary to run everything through `godep` when dependencies are vendored. The Go tool will find them automatically. `godep` should only be a requirement when you need to change/update the dependencies themselves. This is the really good thing about the new vendoring support.

alext:
Github seems to be getting confused by the size of the diffs here.  To repeat the inline comment that seems to have been lost:

It shouldn't be necessary to run everything through `godep` (in the `Makefile`) when dependencies are vendored. The Go tool will find them automatically. `godep` should only be a requirement when you need to change/update the dependencies themselves. This is the really good thing about the new vendoring support.

timblair:
Thanks @alext. I saw the inline comment too :+1:

PR #109, opened 2016-07-29
Fix command for installing godep in readme

rubenarakelyan:
The readme file says to run `go install github.com/tools/godep` to install godep. However, running this in my development VM gave the following error message:

```
can't load package: package github.com/tools/godep: cannot find package "github.com/tools/godep" in any of:
	/usr/lib/goenv/versions/1.6.3/src/github.com/tools/godep (from $GOROOT)
	/var/govuk/gopath/src/github.com/tools/godep (from $GOPATH)
```

Looking at https://github.com/tools/godep/blob/master/Readme.md, the command it suggests is `go get github.com/tools/godep`, which works.

Is this simply a change in naming in newer versions of the `go` command, or am I missing something here?

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ This project uses [Godep][godep] to manage it's dependencies.  If you have a
 working [Go][go] development setup, you should be able to install
 [Godep][godep] by running:
 
-    go install github.com/tools/godep
+    go get github.com/tools/godep
 
 [godep]: https://github.com/tools/godep
 [go]: http://golang.org


alext:
`go install` is for working locally, so was never really the right thing in this instance.  This SO answer explains it well: http://stackoverflow.com/questions/24878737/what-is-the-difference-between-go-get-and-go-install

rubenarakelyan:
Ah great, that makes sense given the original error message. Are you happy to review @alext ?

PR #113, opened 2016-09-05
Update README to clarify local dev requirements

timblair:
Add information to the README to clarify that the router needs to be built from within a Go workspace.

Also remove an unnecessary build target.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,25 +35,21 @@ And some features that we have no need to implement:
 - Response rewriting
 - Authentication
 
-Dependencies
-------------
+Local Setup & Build
+-------------------
 
-This project uses [Godep][godep] to manage it's dependencies.  If you have a
-working [Go][go] development setup, you should be able to install
-[Godep][godep] by running:
+The router needs to be built from within a Go [workspace][workspace] for it to
+correctly identify the vendored dependencies.  Assuming you already have Go and
+a `$GOPATH` set up, you can simply run `go get github.com/alphagov/router` to
+pull the source into your workspace.
 
-    go get github.com/tools/godep
-
-[godep]: https://github.com/tools/godep
-[go]: http://golang.org
-
-Build
------
-
-Once you have Godep installed, you should be able to run:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -35,26 +35,22 @@ And some features that we have no need to implement:
...
-
-    go get github.com/tools/godep
+Local Setup & Build
+-------------------
 
-[godep]: https://github.com/tools/godep
-[go]: http://golang.org
+The router needs to be built from within a Go [workspace][workspace] for it to
+correctly identify the vendored dependencies.  Assuming you already have Go and
+a `$GOPATH` set up, you can simply run `go get github.com/alphagov/repo` to
```
36degrees: `alphagov/router`?

README.md
```diff
@@ -78,6 +74,23 @@ These require a local MongoDB instance and can be run with:
...
+------------
+
+This project uses [Godep][godep] to manage its dependencies.  If you have a
+working [Go][go] development setup, you should be able to install Godep by
+running:
+
+    go get github.com/tools/godep
+
+Note that you only need to install Godep if you need to update or change any
+dependencies.  It's not required to just build or run the application, as all
```
36degrees: 👍 Good to clarify this, thanks.

gpeng:
This is clear. My only concern is that in encouraging using a local workspace is that there may still be the presumption that the router version running on the dev VM will be in sync. I still think having a local $GOPATH within whatever directory is NFS-ed into the dev VM is what we should encourage and then either use some sort of `direnv` thing or have the multiple $GOPATH setting in the article you posted earlier.

gpeng:
That one -> https://dmitri.shuralyov.com/blog/18

timblair:
I don't think we should be giving environment-specific instructions within a given repo's README.  In this case: we shouldn't be giving setup instructions that are specific to the GOV.UK dev VM (that no one else can even see, let alone use): that information is better kept with the realm of the VM and relevant documentation.

gpeng:
Good point, well made :+1:

PR #114, opened 2016-09-07
Upgrade to Go 1.7.1

timblair:


alext:
Would it not be better to upgrade to 1.7.1? I'm thinking https://github.com/golang/go/issues/16865 particularly might be something that could affect us.

timblair:
Yes: I created this PR yesterday, before 1.7.1 was released.  I have a local change to bump to 1.7.1, but I'm waiting on alphagov/ci-puppet#447 being pushed out to CI.

PR #116, opened 2016-11-22
Add Dockerfile and .dockerignore

carlosmartinez:
The new `publishing-e2e-docker` project requires each dependent project
to have a Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+FROM golang:1.7
+
+ADD . /go/src/github.com/alphagov/router
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM golang:1.7
+
+ADD . /go/src/github.com/alphagov/router
+
+RUN go get github.com/tools/godep
```
carlosmartinez: Thanks - I'll see if it works without it.

PR #118, opened 2016-12-06
Router CI scripts

djsd123:
Replaced router bash scripts with groovy scripts for Jenkins

The Jenkinsfile is as generic as possible and calls a router specific groovy script

The Jenkinsfile I feel, could benefit from implementing concurrency as it handles exceptions.


surminus:
Why do we break out the steps into a separate groovy script? For the steps it's taking it looks like it could just be contained into the Jenkinsfile itself. Is there a specific reason we add this additional script?

djsd123:
The same reason they were separate, to begin with, I think.  Reduces coupling and complexity.  We can modify the build of the app completely separate to branch and tag handling.  I feel it serves the purpose and like https://github.com/alphagov/router-api/pull/93, we win having just one Jenkins job.  

We get stung a lot by merging things.

PR #121, opened 2017-02-15
Development defaults in Dockerfile

kevindew:
This allows more of the configuration for docker in this app to be done
closer to the code rather than in various docker-compose files.


kevindew:
thanks for all the reviews @cbaines 👍

PR #126, opened 2017-06-30
[WIP] SSL mongo

SamLR:
This enables SSL connections to Mongo DB by appending "?ssl=true" to the database URL.

### DIFF
--- a/integration_tests/route_helpers.go
+++ b/integration_tests/route_helpers.go
@@ -1,8 +1,13 @@
 package integration
 
 import (
+	"crypto/tls"
+	"fmt"
 	"gopkg.in/mgo.v2"
 	"gopkg.in/mgo.v2/bson"
+	"log"
+	"net"
+	"strings"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -74,10 +79,7 @@ func NewGoneRoute(extraParams ...string) Route {
 }
 
 func init() {
-	sess, err := mgo.Dial("localhost")
-	if err != nil {
-		panic("Failed to connect to mongo: " + err.Error())
-	}
+	sess := getMongoSession("localhost?ssl=true")
 	routerDB = sess.DB("router_test")
 }
 
@@ -88,7 +90,6 @@ func addBackend(id, url string) {
 
 func addRoute(path string, route Route) {
 	route.IncomingPath = path
-
 	err := routerDB.C("routes").Insert(route)
 	Expect(err).To(BeNil())
 }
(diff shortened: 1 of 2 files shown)


surminus:
May be worth getting a review from someone more comfortable with Go /cc @gpeng @timblair

integration_tests/route_helpers.go
```diff
@@ -97,3 +105,36 @@ func clearRoutes() {
 	routerDB.C("routes").DropCollection()
 	routerDB.C("backends").DropCollection()
 }
+
+func getMongoSession(uri string) *mgo.Session {
```
gpeng: Could we extract this out into its own package to reduce duplication between here and `router.go`? That would also allow us to unit test it. I think this would probably be a bit easier to follow as a few shorter methods.

timblair: Agreed. It might also be nice to separate out the creation of the `mgo.DialInfo` value, and the creation of the connection _using_ that value. Doing that will also help with testability: you can test the various permutations of things that will manipulate the `DialInfo` value independently of testing that creation of a connection with _any_ `DialInfo` value.

integration_tests/route_helpers.go
```diff
@@ -97,3 +105,36 @@ func clearRoutes() {
...
+	const sslSuffix string = "?ssl=true"
```
timblair: Do we ever specify any other query string parameters within the connection string? This is a very specific mechanism. We could handle it better by parsing the URL ourselves (error handling omitted for brevity):

```go
// Create the DialInfo value, without SSL info
dialInfo, _ := mgo.ParseURL(uri)

// Manually parse we we can work out if it's an SSL connection
url, _ := url.Parse(uri)

// Check for the relevant SSL parameter and add the DialServer func if necessary
query := url.Query()
if values, hasSSL := query["ssl"]; if hasSSL {
    if len(values) > 0 {
        ssl, _ := strconv.ParseBool(values[0])
        if ssl {
            dialInfo.DialServer = func(...) { ... }
        }
    }
}

// Do the Dialling (but see other comment about separating config from call)
session, _ := mgo.DialWithInfo(dialInfo)
return session
```

integration_tests/route_helpers.go
```diff
@@ -97,3 +105,36 @@ func clearRoutes() {
...
+			log.Fatal(fmt.Println("Failed to connect: %s", err))
+		}
+		return session
+	}
+
+	uri = strings.TrimSuffix(uri, sslSuffix)
+	tlsConfig := &tls.Config{}
+	tlsConfig.InsecureSkipVerify = true
+
+	dialInfo, err := mgo.ParseURL(uri)
```
timblair: I swear I looked for a `mgo.ParseURL` function or similar last week, but didn't find it. ¯\_(ツ)_/¯

deanwilson:
This has been idle too long, and Sam has moved on to a better place, so we're closing the PR.

PR #127, opened 2017-08-15
Push the binary to s3

SamLR:
As part of the build process push the binary to S3 for easier access.


Jenkinsfile
```diff
@@ -61,6 +61,13 @@ node ('mongodb-2.4') {
       govuk.pushTag(REPOSITORY, env.BRANCH_NAME, 'release_' + env.BUILD_NUMBER)
     }
 
+    if (env.BRANCH_NAME == "master") {
+      stage("Push binary to S3") {
+        target_tag = govuk.getNewStyleReleaseTag()
```
surminus: does this mean we're using new tags for these builds? I'm wary of adding new style tags before they were agreed with developers.

SamLR: Good point, I can roll back to using the old (or push both).

PR #129, opened 2017-10-06
Upgrade to Go 1.9.1

gpeng:
Upgrades router to golang 1.9.1. This requires puppet changes to install 1.9.1 on CI before it can be merged.


surminus:
Please can we also make the change in the Dockerfile as we run router as a container in AWS

PR #130, opened 2017-11-27
Push the built Go binary to S3, so we can deploy from AWS

issyl0:
- In PR 128, changes similar to this were reverted because we didn't want to
  use new release tags without consulting with developers. We newly need these
  changes, as we're reverting from Router being deployed on AWS in a container,
  so use the old style release tags (hence why this isn't a simple revert of a
  revert).

https://trello.com/c/IbnMEmul/793-router-move-away-from-container-deployment

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,7 +10,7 @@ node ('mongodb-2.4') {
   env.GOPATH    = "${WORKSPACE}/${BUILD_DIR}"
   env.SRC_PATH  = "${env.GOPATH}/src/github.com/${REPO}"
 
-    def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -61,6 +61,14 @@ node ('mongodb-2.4') {
       govuk.pushTag(REPOSITORY, env.BRANCH_NAME, 'release_' + env.BUILD_NUMBER)
     }
 
+    // Push the Go binary for the build to S3, for AWS releases
```
issyl0: Do we actually want this before line 53 ("archive binaries from the build"), otherwise it won't know what the 'router` artefact is?

surminus: I think the `Archive Artifact` stage creates the artefact, and this stage uploads it to S3, so seems like the correct order to me.

issyl0: Ohhh. I thought it meant "delete/hide", for some reason.

PR #137, opened 2019-08-15
Update MongoDB client, fix racey test, include RDS CA, reduce Docker build size (aka: DocumentDB Support)

chrisfarms:
## What

Tweaks towards a DocumentDB compatible router and improvements to the Docker build that surfaced while testing

* The build docker image was crazy big (nearly 800MB!!) so this switches
to using a multi-stage build and an alpine container to only include the
relevent executable in the image taking it down to a more respectable
12MB.
* We also want to be able to use router with a AWS DocumentDB backend
(mongodb compatible) to do this we need to have the AWS provided CA used
to verify tls connections. So this fetches the CA and places it at
/etc/ssl/certs in the Docker build where golang expects to find it.
* MongoDB client `mgo` was very old (in fact not maintained) ... this was swapped out for a more up to date version that supports TLS connections (required for DocumentDB)
* Dep wasn't playing very nicely with updating the dependancies ... go modules is now the prefured method in the community so switched to using `go mod` to vendor deps
* We hit several race conditions in the tests when testing against DocumentDB (since it's a bit slower than a local mongodb instance) ... so started to replace `sleep Xms` guesses with `Eventually` calls to make them less racey

## Why

Because we intend on replacing mongodb with documentdb


cbaines:
Relating to the "switch to go mod and update vendored deps", how come you made this change?

chrisfarms:
@cbaines (1) `go mod` is now the standard built into go, and is significantly more reliable than the others (especially when just doing `go mod vendor`) .... (2) godep wanted to change _lots_ more things than we wanted as side effect of updating the mongo library, which would have potentially introduced more changes than I felt comfortable with, so rather than fighting it I just ran `go mod vendor` to get the result I wanted.

...also one less developer tool to install

cbaines:
> @cbaines (1) `go mod` is now the standard built into go, and is significantly more reliable than the others (especially when just doing `go mod vendor`) .... (2) godep wanted to change _lots_ more things than we wanted as side effect of updating the mongo library, which would have potentially introduced more changes than I felt comfortable with, so rather than fighting it I just ran `go mod vendor` to get the result I wanted.
> 
> ...also one less developer tool to install

Ok, could you put that in the commit message (for future reference). Thanks.

chrisfarms:
Fixedup commit msg :pen:

PR #141, opened 2019-09-09
Preserve _ga parameter on redirects

bilbof:
We want to preserve the _ga parameter for redirects, even if the route doesn't want other params.

The _ga param will always be passed when redirecting, if it is present.

https://trello.com/c/jBQKyskn/98-preserve-ga-parameter-in-redirects


handlers/redirect_handler.go
```diff
@@ -25,14 +25,22 @@ func addCacheHeaders(writer http.ResponseWriter) {
 	writer.Header().Set("Cache-Control", fmt.Sprintf("max-age=%d, public", cacheDuration/time.Second))
 }
 
+func addGAQueryParam(url string, request *http.Request) string {
+	if ga := request.URL.Query().Get("_ga"); ga != "" {
+		return url + "?_ga=" + ga
```
sihugh: If the destination URL already has a query string (which some do), then this will be a bit wonky

bilbof: Ah, thanks. I forgot about this case.

bilbof:
It seems to be working OK on integration:

https://www-origin.integration.publishing.service.gov.uk/stopcartels?_ga=12345
https://www-origin.integration.publishing.service.gov.uk/contact-consulate-dusseldorf?_ga=hello
https://www-origin.integration.publishing.service.gov.uk/accessibility-regulations?_ga=1235

PR #144, opened 2019-11-21
Close request body on error

1pretz1:
Tests a change to router whereby if an error is thrown for a request, 
the response body is closed.

This comes off the back of a hypothesis from @tlwr on what could be
causing the increased amount of open files on the cache machines.

The increased file handles only seem to occur in production, but we have 
deployed the branch changes to staging to see if any unusual behaviour occurs
before 'testing' this in prod.

Co-authored: @bilbof 

Related incident:
https://docs.google.com/document/d/1gCCSj-WgtK0_TY19J00qf4PiNV5z0KFoLArWkPBgGAg/edit#

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,5 @@ require (
 	golang.org/x/net v0.0.0-20160717224510-3797cd886499 // indirect
 	golang.org/x/sys v0.0.0-20160717071931-a646d33e2ee3 // indirect
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
-	gopkg.in/mgo.v2 v2.0.0-20160609180028-29cc868a5ca6 // indirect
 	gopkg.in/yaml.v2 v2.0.0-20160715033755-e4d366fc3c79 // indirect
 )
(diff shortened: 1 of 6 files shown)


1pretz1:
Does this look okay to you @tlwr? My go knowledge is like zero! 😆

tlwr:
Hi, I am reviewing this.

I don't fully understand the edge cases of this application, so I have written some tests so I can make correct assertions about the current behaviour vs my expectations

Do you mind if I add these tests to the PR?

tlwr:
So I had a look at this and wrote some tests.

Because we're using `httputil.NewSingleHostReverseProxy` we don't actually have to ensure the body is closed ourselves. In fact, closing the body results in a panic:

https://golang.org/src/net/http/httputil/reverseproxy.go#291

I am sorry for sending you down a rabbit hole. You code would have been perfect had I not been wrong...

I'm going to push up my tests and have a think

tlwr:
Awesome, gophers all around! 
image

1pretz1:
Thankyou @tlwr for looking into this and adding tests, looks great! & yup we plan to deploy it Monday'ish! 😄

PR #145, opened 2019-11-27
Improvements to backend handlers

tlwr:
What
----

`DialContext` should be used instead of `Dial` in `http.Transport` and will be removed at a later date. This PR uses `DialContext` but doesn't actually do anything with the context.

`backend_handlers.go` has its own `http.Transport` which is wrapped up in `backendTransport`. By creating the transport directly we do not get any of the defaults from `http.DefaultTransport`. 

I've added some of the configuration parameters from `http.DefaultTransport` to the custom transport, as well as some choices of my own:

- Set `MaxIdleConns` to 60, in the default transport this is 100, but in a new transport, it defaults to zero, which means there is no limit to idle connections. I have set this to 60 so that the app does keep lots of idle connections open.

- Set `IdleConnTimeout`, in the default transport this is 90 seconds, in a new transport there is no limit, so the app would keep an idle connection open forever, even if the upstream closed it. This is a resource leak. I have set it to 10 minutes, which is lower than the idle timeout of GOV.UK PaaS which is 15 minutes.

- Set `ExpectContinueTimeout` and `TLSHandshakeTimeout` to the default transport values.

I also ran `gofmt` and added a Make task to help run the tests locally.

How to review
----

Run the tests

Code review


handlers/backend_handler.go
```diff
@@ -56,22 +57,47 @@ type backendTransport struct {
 // Construct a backendTransport that wraps an http.Transport and implements http.RoundTripper.
 // This allows us to intercept the response from the backend and modify it before it's copied
 // back to the client.
-func newBackendTransport(connectTimeout, headerTimeout time.Duration, logger logger.Logger) (transport *backendTransport) {
-	transport = &backendTransport{&http.Transport{}, logger}
+func newBackendTransport(connectTimeout, headerTimeout time.Duration, logger logger.Logger) *backendTransport {
```
chrisfarms: Is there a reason not to consume the context? ... think if you create your own net.Dailer you can use the DialContext method to honour the context... but not sure if it is intentional or not

```
dialer := &net.Dialer{Timeout: connectTimeout}
transport.DialContext = func(ctx context.Context, network, address string) (net.Conn, error) {
	return dialer.DialContext(ctx, network, address)
}
```

tlwr: You have a very good point

I did it this way because ignoring the context is equivalent to the behaviour before, just using a non-deprecated function.

I guess what would be better is your solution but also configuring the `Dialer` to have good defaults other than the Timeout:

```
    DialContext: (&net.Dialer{
        Timeout:   connectTimeout,   // Configured by caller
        KeepAlive: 30 * time.Second, // same as DefaultTransport
        DualStack: true,             // same as DefaultTransport
    }).DialContext,
```

1pretz1:
Toby and I did some investigation on staging and production.

The router app runs via the `deploy` user:

```
$ su deploy

$ ss -tp | grep router
# lots of open TCP connections in ESTAB

$ ss -tp | grep router | awk '{print $5 " " $6}' | column -t  | grep -v '^10\.' | grep -v '^127\.'
# all outbound connections

$ ss -tp | grep router | awk '{print $5 " " $6}' | column -t  | grep -v '^10\.' | grep -v '^127\.' | wc -l 
# number of file descriptors
336

$ ss -tp | grep router | awk '{print $5 " " $6}' | column -t  | grep -v '^10\.' | grep -v '^127\.' | cut -d, -f3 | tr -d '()' | xargs -I {} stat /proc/20130/fd/{} | grep Change | awk '{print $2}' | sort | uniq -c
# file descriptors by change date
    256 2019-11-28
     80 2019-11-29
```

It seems the router app is holding connections open basically forever, so we should probably explicitly set a maximum idle timeout.

PR #147, opened 2020-02-28
[spike] Return a 405 if routes does not allow a method

bilbof:
**Note**: This is not yet ready for merge.

This adds optional methods to routes, allowing us to specify the routes that should be accepted by a path.

For example, on GOV.UK 'GET /' is acceptable but 'DELETE /' is not. We know this, so we can return a 405 at the router level, rather than letting our slower applications handle this request downstream of router.

Further work will need to be done, first on Router API to support path methods, then on content store, publishing-api and the publishing apps.

Router API should set a default empty array for all routes, or nil, which will permit all methods to be accepted. Router API should also allow HEAD should a GET method be allowed.

I'd be grateful for feedback on the router implementation, and thoughts about the approach more generally. Let me know if you have any questions!

https://trello.com/c/v9exwjls/1406-spike-implement-request-method-handling-in-router

----

@kevindew here adding some adding context for this work. The work Bill has put together is a spike exploring the hypothesis that for requests with the wrong HTTP method it would be a lower performance cost to have them returned directly from Router rather than having the request be proxied to the app to then return a 404/405 response. This is an alternative approach to [caching 404/405 POST requests at the CDN](https://github.com/alphagov/govuk-cdn-config/pull/160) which suffered the problem that cache entries were unable to be purged and felt, to us, like an unconventional usage of varnish.

The idea we had was that we could [specify in Router API which methods are accepted](https://github.com/alphagov/router-api/commit/9fa4f3d31694ae95da25987f5609462bd32281ab) for a route, with GET being the default if no method provided and HEAD implicitly supported wherever GET is available. The expectation is then that whenever GOV.UK apps register routes they could specify if they wanted routes beyond GET. This would involve a [schema change](https://github.com/alphagov/govuk-content-schemas/commit/53f68f27d9c830aef32ad6ea930a66a402a6b42e) and [Content Store passing the methods to Router API](https://github.com/alphagov/content-store/commit/f50908c9cdd10402171bb11e88b5537113698510).

The next step would be to then change all of the GOV.UK routes that accept non-GET requests (which, as far as I can tell, is only POST requests) [to also accept POST requests](https://github.com/alphagov/email-alert-frontend/commit/2c1a04fd790e3cb1c84cc5d2e65fd32b582ff63d). I think I managed to identify most from CDN logs and config/routes.rb files, however there were more than I anticipated, more prefix ones than expected and I'm not super confident I caught them all.

- Email Alert Frontend sets it's own POST prefix paths of /email-signup, /email/manage, /email/subscriptions and /email/unsubscribe. It also serves /*base-path/email-signup paths set by the following apps: [Search API](https://github.com/alphagov/search-api/blob/00a56065dac50d377c9bcd76b233e605f08cc868/lib/tasks/publishing_api.rake#L48), [Travel Advice Publisher](https://github.com/alphagov/travel-advice-publisher/blob/e291b61198f0587887438e72f654181068ddbae5/lib/tasks/publishing_api.rake#L103-L115), [Specialist Publisher](https://github.com/alphagov/specialist-publisher/blob/0e12ff7e36765719294e725129d2c1e1abef62a4/lib/publishing_api_finder_publisher.rb#L55-L58), [Service Manual Publisher](https://github.com/alphagov/service-manual-publisher/blob/432dc8072ffe33578b6ce43039fc9b97adf0bea3/app/presenters/service_standard_email_alert_signup_presenter.rb#L26-L31) and [Special Route Publisher](https://github.com/alphagov/special-route-publisher/blob/0e0bfa8fdf9aef998a86601ac787c87aadc173f0/data/special_routes.yaml#L18-L21)
- Frontend has [/find-local-council](https://github.com/alphagov/frontend/blob/0dcd33bb994e743eab94f34911656163c3678258/lib/special_route_publisher.rb#L21). It also has a lot of [/brexit-funding URLs](https://github.com/alphagov/frontend/blob/0dcd33bb994e743eab94f34911656163c3678258/config/routes.rb#L26-L46) which may no longer be in use.
- Publisher would need to set POST routes for [local transaction, place and licence document types](https://github.com/alphagov/frontend/blob/0dcd33bb994e743eab94f34911656163c3678258/config/routes.rb#L66-L85). It also sets a [/*path/service-sign-in](https://github.com/alphagov/publisher/blob/404862d6fdcd2b5ae939dc13441215a9b4bd066d/app/presenters/formats/service_sign_in_presenter.rb#L58) which [government-frontend accepts as a POST](https://github.com/alphagov/government-frontend/blob/7bd03d3e521c373464758bbcf19d3f94ae526bf6/config/routes.rb#L25)
- Contacts publishes [routes that accept POST on /contact/govuk/*](https://github.com/alphagov/feedback/blob/eab6a4659e1ad711f83e3ce39b7c89e3ad59c389/lib/tasks/publishing_api.rake#L63-L70)
- Licensify seems to have a /apply-for-a-licence prefix route that isn't registered in Content Store

Once these are all changed then the default of GET could be set. 

Bill and I have decided to pause work on this as we felt the next step would be an RFC and we struggled to justify the value of this work relative to the effort. We felt it was perhaps too niche a problem to solve. If this is picked up later we think the next step is do some load testing with router to test the hypothesis that this will allow a much greater number of requests to be served with Router better equipped to determine if they are erroneous without proxying.

### DIFF
--- a/triemux/mux.go
+++ b/triemux/mux.go
@@ -9,6 +9,7 @@ import (
 	"hash"
 	"log"
 	"net/http"
+	"sort"
 	"strings"
 	"sync"
 )
@@ -24,6 +25,12 @@ type Mux struct {
 type muxEntry struct {
 	prefix  bool
 	handler http.Handler
+	methods []string
+}
+
+type muxError struct {
+	code    int
+	message string
 }
 
 // NewMux makes a new empty Mux.
@@ -41,18 +48,18 @@ func (mux *Mux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	handler, ok := mux.lookup(r.URL.Path)
-	if !ok {
-		http.NotFound(w, r)
+	handler, err := mux.lookup(r.Method, r.URL.Path)
+	if err != nil {
+		http.Error(w, err.message, err.code)
 		return
 	}
 
 	handler.ServeHTTP(w, r)
 }
 
-// lookup takes a path and looks up its registered entry in the mux trie,
+// lookup takes a method and path and looks up its registered entry in the mux trie,
 // returning the handler for that path, if any matches.
-func (mux *Mux) lookup(path string) (handler http.Handler, ok bool) {
+func (mux *Mux) lookup(method string, path string) (http.Handler, *muxError) {
 	mux.mu.RLock()
 	defer mux.mu.RUnlock()
 
@@ -61,35 +68,43 @@ func (mux *Mux) lookup(path string) (handler http.Handler, ok bool) {
 	if !ok {
 		val, ok = mux.prefixTrie.GetLongestPrefix(pathSegments)
 	}
+
 	if !ok {
-		return nil, false
+		return nil, &muxError{ code: http.StatusNotFound, message: "404 Not Found" }
 	}
 
 	entry, ok := val.(muxEntry)
 	if !ok {
 		log.Printf("lookup: got value (%v) from trie that wasn't a muxEntry!", val)
-		return nil, false
+		return nil, &muxError{ code: http.StatusInternalServerError, message: "500 Internal Server Error" }
 	}
 
-	return entry.handler, ok
+	if !entry.methodSupported(method) {
+		return nil, &muxError{ code: http.StatusMethodNotAllowed, message: "405 Method Not Allowed" }
+	}
+
+	return entry.handler, nil
 }
 
 // Handle registers the specified route (either an exact or a prefix route)
 // and associates it with the specified handler. Requests through the mux for
 // paths matching the route will be passed to that handler.
-func (mux *Mux) Handle(path string, prefix bool, handler http.Handler) {
+func (mux *Mux) Handle(path string, methods []string, prefix bool, handler http.Handler) {
 	mux.mu.Lock()
 	defer mux.mu.Unlock()
 
-	mux.addToStats(path, prefix)
+	// For faster lookups and consistent checksums
+	sort.Strings(methods)
+
+	mux.addToStats(path, prefix, methods)
 	if prefix {
-		mux.prefixTrie.Set(splitpath(path), muxEntry{prefix, handler})
+		mux.prefixTrie.Set(splitpath(path), muxEntry{prefix, handler, methods})
 	} else {
-		mux.exactTrie.Set(splitpath(path), muxEntry{prefix, handler})
+		mux.exactTrie.Set(splitpath(path), muxEntry{prefix, handler, methods})
 	}
 }
 
-func (mux *Mux) addToStats(path string, prefix bool) {
+func (mux *Mux) addToStats(path string, prefix bool, methods []string) {
 	mux.count++
 	mux.checksum.Write([]byte(path))
 	if prefix {
@@ -97,6 +112,7 @@ func (mux *Mux) addToStats(path string, prefix bool) {
 	} else {
 		mux.checksum.Write([]byte("(false)"))
 	}
+	mux.checksum.Write([]byte(strings.Join(methods,",")))
 }
 
 func (mux *Mux) RouteCount() int {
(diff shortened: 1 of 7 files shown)


triemux/mux.go
```diff
@@ -107,6 +123,21 @@ func (mux *Mux) RouteChecksum() []byte {
 	return mux.checksum.Sum(nil)
 }
 
+func (e muxEntry) methodSupported(method string) bool {
+	if len(e.methods) == 0 {
+		// If routes don't specify the methods they allow, we assume they
+		// allow all methods.
+		return true
+	}
```
kevindew: Nice 👍

triemux/mux.go
```diff
@@ -107,6 +123,21 @@ func (mux *Mux) RouteChecksum() []byte {
...
+	return contains(e.methods, method)
```
kevindew: I think we'd also want a condition in here that returns true if we're given a HEAD and GET is a method. I don't think we'd want every GET to be registered with a corresponding HEAD

triemux/mux.go
```diff
@@ -61,42 +68,51 @@ func (mux *Mux) lookup(path string) (handler http.Handler, ok bool) {
...
 	if !ok {
-		return nil, false
+		return nil, &muxError{ code: http.StatusNotFound, message: "404 Not Found" }
 	}
 
 	entry, ok := val.(muxEntry)
 	if !ok {
 		log.Printf("lookup: got value (%v) from trie that wasn't a muxEntry!", val)
-		return nil, false
+		return nil, &muxError{ code: http.StatusInternalServerError, message: "500 Internal Server Error" }
```
kevindew: Oh this would make more sense than the previous implementation.

triemux/mux.go
```diff
@@ -61,42 +68,51 @@ func (mux *Mux) lookup(path string) (handler http.Handler, ok bool) {
...
+	}
+
+	if !entry.methodSupported(method) {
+		return nil, &muxError{ code: http.StatusMethodNotAllowed, message: "405 Method Not Allowed" }
```
kevindew: Aces

tlwr:
_I don't work on a GOV.UK team (just within RE) so please ignore these comments if they are not relevant_

---

⬆️ Kevin's comment about `GET` and `HEAD`; 👍 

🏅  I like the request method sorting when calling `Handle`. Elegant

❓ Overlapping routes currently reject each other, [commit adding failing test where reading from backend-1 fails](https://github.com/alphagov/router/commit/60a58359378672c41b9e2abccea74b94e184e64a). This is may be desired behaviour, but worth double checking.

❓ How do you feel about supporting [the method `OPTIONS`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS). Otherwise this may break sites in _interesting_ ways (pre-flight). Because router now knows about what methods are registered for each routes:
- When methods for route is `[]`, then return `*`
- When methods for route is not `[]` then return `OPTIONS` and the methods specified on the route

❓ How do you feel about [the `Allow` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Allow). When returning a `405`, it is good practice to return what methods are allowed for a response.

bilbof:
Thank you so much @tlwr, this is fantastic feedback 👍 

> ❓ Overlapping routes currently reject each other, commit adding failing test where reading from backend-1 fails. This is may be desired behaviour, but worth double checking.

Great spot @tlwr! I wonder whether we'd want to handle this logic in router-api? In router-api [route paths have a unique constraint](https://github.com/alphagov/router-api/blob/master/app/models/route.rb#L14) so we wouldn't see this error at the moment.

> ❓ How do you feel about supporting the method OPTIONS.

That's super interesting, we could definitely add support for this. It might even make all pre-flight requests a bit faster!

> ❓ How do you feel about the Allow header. When returning a 405, it is good practice to return what methods are allowed for a request.

Yep we should do this too!

triemux/mux.go
bilbof: Yep that should be easy to handle in router.

bilbof:
@kevindew thanks for your review!

> forgiving if routes were defined with lower case strings

Yeah it might be a weird bug to investigate if router-api doesn't validate that the provided methods are acceptable and uppercase! Perhaps we could start strict and add validations in router api and content schemas to enforce methods to be uppercase; it'd be easier to become more lenient in future if necessary.

If it's OK with you, I'll make the changes you and Toby suggested after we meet up on Thursday and determine whether we want to continue with this spike?

kevindew:
> @kevindew thanks for your review!
> 
> > forgiving if routes were defined with lower case strings
> 
> Yeah it might be a weird bug to investigate if router-api doesn't validate that the provided methods are acceptable and uppercase! Perhaps we could start strict and add validations in router api and content schemas to enforce methods to be uppercase; it'd be easier to become more lenient in future if necessary.
> 
> If it's OK with you, I'll make the changes you and Toby suggested after we meet up on Thursday and determine whether we want to continue with this spike?

Yup that sounds good - I've found a bunch of information out too for us to discuss.

kevindew:
@bilbof I've added context and my investigation links to the description of this PR.

PR #149, opened 2020-03-04
Adds a test to demonstrate encoded path matching

sihugh:
This is a copy / tweak of existing tests and demonstrates that router routes things that are URL encoded.

At least one of our frontend apps is not built to handle URL encoded paths and [errors under some circumstances](https://sentry.io/organizations/govuk/issues/1399727649/?project=202213&query=is%3Aunresolved).  [A sort of fix has been implemented for collections](https://github.com/alphagov/collections/pull/1533).

Should we "fix" router to not behave like this?

Related stack overflows are seem inconclusive - for example Apache allows you to toggle a setting that allows this. 
- https://stackoverflow.com/questions/3235219/urlencoded-forward-slash-is-breaking-url
- https://stackoverflow.com/questions/1957115/is-a-slash-equivalent-to-an-encoded-slash-2f-in-the-path-portion-of-a?

### DIFF
--- a/integration_tests/route_selection_test.go
+++ b/integration_tests/route_selection_test.go
@@ -155,6 +155,23 @@ var _ = Describe("Route selection", func() {
 			})
 		})
 
+		Describe("url encoded paths", func() {
+			BeforeEach(func() {
+				addRoute("/foo/bar", NewBackendRoute("inner-backend"))
+				reloadRoutes()
+			})
+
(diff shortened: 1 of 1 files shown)


sengi:
Hmm interesting. Sounds like the current behaviour has potential for [request smuggling](https://cwe.mitre.org/data/definitions/444.html) shenaningans (but I've spent like 30 s thinking about it so not making any claims here!)

PR #161, opened 2020-09-21
Show code coverage after running tests

benthorner:
https://trello.com/c/epNWudCR/176-write-an-rfc-for-continuous-deployment-and-seek-review

We should always monitor the amount of code that is exercised when
we run our tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ build:
 
 test: build
 	go test -race ./trie ./triemux ./handlers
-	go test -race -v ./integration_tests
+	go test -cover -race -v ./integration_tests
 
 run: build
 	$(BINARY)


Makefile
```diff
@@ -18,7 +18,7 @@ build:
 
 test: build
 	go test -race ./trie ./triemux ./handlers
-	go test -race -v ./integration_tests
+	go test -cover -race -v ./integration_tests
```
kevindew, line 21: Do we not need a coverage output on the previous command too?

benthorner: Hmm...this is proving to be more difficult than I thought - thanks for questioning. It turns out that the `-cover` option only looks at files in the same package, which is unhelpful since these tests cover all packages and also [operate on the compiled binary](https://github.com/alphagov/router/blob/master/integration_tests/router_support.go#L48). It looks like there might be [a hack around it](https://blog.cloudflare.com/go-coverage-with-external-tests/), which may suffice to get a one-time coverage report for this repo. I'll close this PR, since it's not clearly possible to achieve the stated outcome.

benthorner:
Closing (see https://github.com/alphagov/router/pull/161#discussion_r503235260).

Makefile
benthorner: Some more comments for anyone looking at this in future: [although this problem may eventually be solved](https://github.com/golang/go/issues/51430), for now if we want to check the coverage of this repo I think we'll need to do it manually. That's not as hard as it sounds. As a starting point:

```
go test -coverprofile unit_test_coverage.json -race $(go list ./... | grep -v integration_tests)
go tool cover -html=unit_test_coverage.json
```

This gives an initial picture of what code might not be covered. The next step would be to look at which bits of it might actually be covered by the integration tests. Rather than worrying about a percentage, focus on the individual features and edge cases that might not be covered and make a prioritised list of them all; put a divide in the list to establish which features and edge cases we need to ensure are covered e.g.

- router.go: `NewRouter` (happy path)
- ...
- router.go: `NewRouter` L66: `if err != nil` (invalid poll interval)

You can then put a unique "print" statement in each place and run the integration tests to see which of them are actually covered and cross those that are off the list.

There could be a smarter approach, but this is at least a starting point.

PR #188, opened 2021-02-22
Enable Router to be configured with multiple domains

bilbof:
## Problem

During the migration to ECS, Replatforming intends to run router in ECS and EC2, using the same backing database managed by Router API, hosted in EC2.

One problem with this approach is that ECS and EC2 apps are served at different domains (as below):

- https://calendars.integration.govuk-internal.digital/
- http://calendars.mesh.govuk-internal.digital/

Router looks up the entire backend_url from its database. This currently makes it impossible to run two different Router applications which serve different domains, but use the same database.

## Solution

Rather than store the backend_url for a backend in Router API's database, we'll provide backend URLs as environment variables such as

```
BACKEND_URL_content-store=https://content-store.gov.uk/
BACKEND_URL_search-api=https://search.gov.uk/
```

If an environment variable matching a backend_id is set Router will use the value of the variable instead of the FQDN backend_url stored in the DB.

This change should have no impact on the current platform since we won't set any new environment variables there.

As it is trivial to bring up new containers in ECS with different sets of environment variables, this will enable us to make router work across domains while using the same database (for workspaces).

Migration path:

1. Push this change out
2. Add env var for backends to router instances in ECS

**Other approaches**

* Provide an env var for the host domain and add subdomains to backends in Router API (overly complex)
* write duplicate backend records to the db (unnecessary)

Trello: https://trello.com/c/qtAZfkjg/403-work-out-whether-we-can-make-router-work-with-multiple-domains-and-a-single-database
Design doc: https://docs.google.com/document/d/147FSn4sd80iGZ1KxcJ7OYG2cW_uY-Gc7e7laLhkkMJo/edit#

### DIFF
--- a/integration_tests/route_loading_test.go
+++ b/integration_tests/route_loading_test.go
@@ -11,6 +11,7 @@ var _ = Describe("loading routes from the db", func() {
 	var (
 		backend1 *httptest.Server
 		backend2 *httptest.Server
+		backend3 *httptest.Server
 	)
 
 	BeforeEach(func() {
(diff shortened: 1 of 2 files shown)


richardTowers:
I think you might need to update the PR description to match the commit too :)

bilbof:
@richardTowers thanks for your feedback! I've improved the test and formatted.

richardTowers (approved):
Beautiful, thanks Bill!

PR #193, opened 2021-03-18
Improve error reporting during restarts

bilbof:
This change is intended to make it easier to debug the behaviour of router during a deployment.

During a deploy we use the [tablecloth](https://github.com/alext/tablecloth) package to bring up a child process while the parent process restarts. We've been seeing 503 errors during this procedure.

The problem this solves is that we don't yet know what proportion of errors are due to downstream (Fastly/Nginx/Varnish) timeouts, or due to router serving a 503 when the in-memory routing table (trie) is empty. We know that Router does serve some 500s, which we think are due to client timeouts. This change should give us the answer.

I've included a couple of bumps of packages (sentry and tablecloth). This change does not intend to fix the underlying problem.


triemux/mux.go
```diff
@@ -37,6 +40,13 @@ func NewMux() *Mux {
 // If the routing table is empty, return a 503.
 func (mux *Mux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if mux.count == 0 {
+		defer func() {
```
richardTowers, line 43: Do you actually need the `defer func()` here? Could you just inline this code after `w.WriteHeader(http.StatusServiceUnavailable)`? Or is there a chance that will panic?

triemux/mux.go
```diff
@@ -37,6 +40,13 @@ func NewMux() *Mux {
...
+			logger.NotifySentry(logger.ReportableError{
+				Error: logger.RecoveredError{"Route table is empty!"},
+				Request: r,
+			})
+			InternalServiceUnavailableCountMetric.With(prometheus.Labels{"host": r.Host}).Inc()
```
richardTowers, line 48: Is `r.Host` always going to be `www.gov.uk`? Is it worth having that label?

bilbof: True, I'll remove it. The host label is probably not useful - I added it since other count metrics have it.

The label for `TEMPORARY_CHILD` should work. I'm concerned that Router seems to be briefly [unreachable during deploys](https://prometheus.publishing.service.gov.uk/graph?g0.expr=up%7Binstance%3D%2210.13.5.208%3A3055%22%7D&g0.tab=0&g0.stacked=0&g0.range_input=20m&g0.end_input=2021-03-16%2017%3A20%3A13&g0.moment_input=2021-03-16%2017%3A20%3A13); I am not sure if we actually scrape the temporary child.

triemux/mux.go
bilbof: True, we don't need to `defer` this.

bilbof:
I've pushed up the suggested changes.

I've written up the results of the investigations on [the tech debt story](https://trello.com/c/FO8tFR3G/149-we-cant-deploy-router-without-causing-requests-to-error#comment-6053cc64ee56ef823190c00e). It's not a high priority since we'll stop using Tablecloth once Router is run in ECS, but I'll share this with the SRE community to see if they have thoughts.

richardTowers (approved):

PR #195, opened 2021-03-22
Update routes within router

karlbaker02:
This PR adds functionality to Router so that it takes charge of updating its own routes, polling MongoDB every 30 seconds to check for changes and performing the swapping of route tables if updates are found to have occurred. This change takes place at the same time as we stop requests being sent by Router API to force updates to Router instances, allowing us to decouple these apps and make it far easier to run in ECS (as Router API no longer needs to know where to find Router instances).

[For more information around these changes, view RFC 135](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-135-updating-routes-in-router.md).

### DIFF
--- a/main.go
+++ b/main.go
@@ -18,6 +18,7 @@ var (
 	apiAddr               = getenvDefault("ROUTER_APIADDR", ":8081")
 	mongoURL              = getenvDefault("ROUTER_MONGO_URL", "127.0.0.1")
 	mongoDbName           = getenvDefault("ROUTER_MONGO_DB", "router")
+	mongoPollInterval     = getenvDefault("ROUTER_MONGO_POLL_INTERVAL", "2s")
 	errorLogFile          = getenvDefault("ROUTER_ERROR_LOG", "STDERR")
 	tlsSkipVerify         = os.Getenv("ROUTER_TLS_SKIP_VERIFY") != ""
 	enableDebugOutput     = os.Getenv("DEBUG") != ""
@@ -32,12 +33,13 @@ Usage: %s [-version]
 
 The following environment variables and defaults are available:
 
-ROUTER_PUBADDR=:8080        Address on which to serve public requests
-ROUTER_APIADDR=:8081        Address on which to receive reload requests
-ROUTER_MONGO_URL=127.0.0.1  Address of mongo cluster (e.g. 'mongo1,mongo2,mongo3')
-ROUTER_MONGO_DB=router      Name of mongo database to use
-ROUTER_ERROR_LOG=STDERR     File to log errors to (in JSON format)
-DEBUG=                      Whether to enable debug output - set to anything to enable
+ROUTER_PUBADDR=:8080             Address on which to serve public requests
+ROUTER_APIADDR=:8081             Address on which to receive reload requests
+ROUTER_MONGO_URL=127.0.0.1       Address of mongo cluster (e.g. 'mongo1,mongo2,mongo3')
+ROUTER_MONGO_DB=router           Name of mongo database to use
+ROUTER_MONGO_POLL_INTERVAL=2s    Interval to poll mongo for route changes
+ROUTER_ERROR_LOG=STDERR          File to log errors to (in JSON format)
+DEBUG=                           Whether to enable debug output - set to anything to enable
 
 Timeouts: (values must be parseable by http://golang.org/pkg/time/#ParseDuration)
 
@@ -105,11 +107,11 @@ func main() {
 		tablecloth.WorkingDir = wd
 	}
 
-	rout, err := NewRouter(mongoURL, mongoDbName, backendConnectTimeout, backendHeaderTimeout, errorLogFile)
+	rout, err := NewRouter(mongoURL, mongoDbName, mongoPollInterval, backendConnectTimeout, backendHeaderTimeout, errorLogFile)
 	if err != nil {
 		log.Fatal(err)
 	}
-	rout.ReloadRoutes()
+	go rout.SelfUpdateRoutes()
 
 	wg := &sync.WaitGroup{}
 	wg.Add(2)
--- a/router.go
+++ b/router.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -14,6 +15,7 @@ import (
 	"github.com/alphagov/router/logger"
 	"github.com/alphagov/router/triemux"
 	"github.com/globalsign/mgo"
+	"github.com/globalsign/mgo/bson"
 )
 
 // Router is a wrapper around an HTTP multiplexer (trie.Mux) which retrieves its
@@ -23,15 +25,28 @@ type Router struct {
 	lock                  sync.RWMutex
 	mongoURL              string
 	mongoDbName           string
+	mongoPollInterval     time.Duration
 	backendConnectTimeout time.Duration
 	backendHeaderTimeout  time.Duration
+	mongoReadToOptime     bson.MongoTimestamp
 	logger                logger.Logger
+	ReloadChan            chan bool
 }
 
 type Backend struct {
 	BackendID     string `bson:"backend_id"`
 	BackendURL    string `bson:"backend_url"`
 	SubdomainName string `bson:"subdomain_name"`
+}
+
+type MongoReplicaSet struct {
+	Members []MongoReplicaSetMember `bson:"members"`
+}
+
+type MongoReplicaSetMember struct {
+	Name    string              `bson:"name"`
+	Optime  bson.MongoTimestamp `bson:"optime"`
+	Current bool                `bson:"self"`
 }
 
 type Route struct {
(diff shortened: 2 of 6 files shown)


karlbaker02:
Not sure how much logging we want to keep from what I've added: all the useful stuff that I used to debug I've made `logDebug` and everything else kept as `logInfo` to have what's hopefully the minimally-viable amount of logs to know when routes are / are not updated. Happy to change this as people see fit.

router.go
```diff
@@ -148,6 +210,46 @@ func (rt *Router) ReloadRoutes() {
...
+		if instance.Current == true {
+			currentInstance = append(currentInstance, instance)
+		}
+	}
+
+	logDebug("router: MongoDB instances", currentInstance)
+
+	if len(currentInstance) != 1 {
+		panic(fmt.Sprintf("router: did not find exactly one current MongoDB instance, skipping update (current instances found: %d)", len(currentInstance)))
+	}
```
sengi, line 248: Thanks for considering this edge case :) I wonder if it might be safer to make it log an error and just return the first one that it found though, rather than not updating at all. What do you reckon?

sengi (commented):
Haven't fully reviewed but spotted a couple of things I thought worth mentioning - hopefully useful.

sengi:
> Not sure how much logging we want to keep from what I've added: all the useful stuff that I used to debug I've made `logDebug` and everything else kept as `logInfo` to have what's hopefully the minimally-viable amount of logs to know when routes are / are not updated. Happy to change this as people see fit.

I think the logging is fine for now, but (just as an optional suggestion for later) you could get better value out of the log lines that you already have by adding specifics to them. For example I would consider either including the timestamps in the `updates found` / `no updates found` messages (so that someone reading the log can tell *why* it decided it found / didn't find an update) or removing them. I agree that Info feels like the right level for those messages, FWIW.

On a related note, it'd be great to add a Prometheus metric for the last-updated timestamp. Not for this PR though - just mentioning as a possibility for later.

router.go
```diff
@@ -63,16 +76,64 @@ func NewRouter(mongoURL, mongoDbName, backendConnectTimeout, backendHeaderTimeou
...
 		mongoURL:              mongoURL,
 		mongoDbName:           mongoDbName,
 		backendConnectTimeout: beConnTimeout,
 		backendHeaderTimeout:  beHeaderTimeout,
+		mongoReadToOptime:     mongoReadToOptime,
 		logger:                l,
+		ReloadChan:            reloadChan,
 	}
+
+	go func(r chan bool) {
```
sengi, line 99: Consider making this a (package-visible) method of Router rather than an nested, anonymous function - it's very much nontrivial and quite important, so I think it deserves a proper declaration.

router.go
karlbaker02: That's a possibility, but if you do this you then open yourself up to other edge cases i.e. you used the timestamp from the instance you were connected to but might then get another set of routes from another instance that hasn't yet updated, meaning you'd be back into the race condition we talked about a while ago. This would be a problem during times of infrequent changes, so probably not highly likely, but worth thinking about.

karlbaker02:
> > Not sure how much logging we want to keep from what I've added: all the useful stuff that I used to debug I've made `logDebug` and everything else kept as `logInfo` to have what's hopefully the minimally-viable amount of logs to know when routes are / are not updated. Happy to change this as people see fit.
> 
> I think the logging is fine for now, but (just as an optional suggestion for later) you could get better value out of the log lines that you already have by adding specifics to them. For example I would consider either including the timestamps in the `updates found` / `no updates found` messages (so that someone reading the log can tell _why_ it decided it found / didn't find an update) or removing them. I agree that Info feels like the right level for those messages, FWIW.
> 
> On a related note, it'd be great to add a Prometheus metric for the last-updated timestamp. Not for this PR though - just mentioning as a possibility for later.

I did think about the specifics, actually timestamp of when the event happened, but then I figured that we already have this information recorded (it must be by the logging functions) in Kibana, as well as timestamps provided by Kibana itself (or at least visible therein).

The Prometheus metric sounds like a good thing.

router.go
karlbaker02: Fair point, though I would counter that with what I'm doing here is no different from the reload function that currently exists within _router_api.go_

router.go
```diff
@@ -148,6 +210,46 @@ func (rt *Router) ReloadRoutes() {
...
+	defer func() {
+		if r := recover(); r != nil {
+			logWarn(r)
+		}
+	}()
+
+	replicaSetStatus := bson.M{}
+
+	if err := db.Run("replSetGetStatus", &replicaSetStatus); err != nil {
+		panic(fmt.Sprintf("router: couldn't get replica set status from MongoDB, skipping update (error: %v)", err))
```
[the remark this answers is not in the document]
karlbaker02: To address your points in turn:

- If I weren't to use `panic` here, I would probably just `logWarn` for each of the errors and then return. We could write a named function to encapsulate this, but that feels too far to go for a simple one-liner to `logWarn`.

- I can definitely see what you're saying about potentially obscuring the logic at the caller. I intentionally didn't return here so as to return a zeroed `MongoReplicaSetMember` for which the `Optime` property would be the default value for an `int64` (that is `0`), to ensure that we can do the optime comparison and know whether to update or not. Happy to change this to make it more golang-esque.

- On the point of the `defer ... recover` on lines 105-109, if I'm not mistaken we are just failing the update (the process will not crash and stop serving traffic). This is in the `for` loop on line 103 which will be entered when we have something placed on to the `reloadChan`, where we then define the anonymous function (from line 104) and immediately call that anonymous function (we do this so as to ensure that all deferred calls are executed when we expect them to be).

router.go
sengi: Not a big deal - feel free to take or leave this one :)

router.go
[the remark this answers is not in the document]
karlbaker02: This is now done

router.go
karlbaker02: This is now done

sengi (approved):
I'm having a bit of crisis of conscience with this review if I'm honest. I'm used to a much higher bar for well-structured code and testing, and the lack of tests for this really worry me. On the other hand, the existing tests are in a poor state to begin with and this is an essentially unowned code base with a lot of pre-existing issues, and I sense that you're feeling pressure to get this out the door.

I'll therefore defer to your judgement as to where to set the bar. As an experienced GOV.UK developer you no doubt have a better sense of where that bar is than I do. So if you're happy to roll this out, then I don't want to block you. I appreciate you've done some manual testing and let the branch soak in Integration for a bit.

(In case you're wondering what I would do if it were my change: I'd probably split up and simplify pollAndReload to make it a more obvious "if shouldReload() then reload" loop, rip out those dubious pre-existing/copied `recover`s and the anonymous functions that are only there because of them, and add some tests for all the stuff I'd touched - which would then be easier/possible to write once the nontrivial stuff is no longer buried in anon functions. Nothing stopping us doing that later, of course.)

main.go
```diff
@@ -36,6 +37,7 @@ ROUTER_PUBADDR=:8080        Address on which to serve public requests
 ROUTER_APIADDR=:8081        Address on which to receive reload requests
 ROUTER_MONGO_URL=127.0.0.1  Address of mongo cluster (e.g. 'mongo1,mongo2,mongo3')
 ROUTER_MONGO_DB=router      Name of mongo database to use
+ROUTER_MONGO_POLL_INTERVAL=2    Interval in seconds to poll mongo for route changes
```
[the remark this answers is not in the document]
karlbaker02: Thanks, the `s` here was actually missing (I've added it elsewhere).

karlbaker02:
> I'm having a bit of crisis of conscience with this review if I'm honest. I'm used to a much higher bar for well-structured code and testing, and the lack of tests for this really worry me. On the other hand, the existing tests are in a poor state to begin with and this is an essentially unowned code base with a lot of pre-existing issues, and I sense that you're feeling pressure to get this out the door.
> 
> I'll therefore defer to your judgement as to where to set the bar. As an experienced GOV.UK developer you no doubt have a better sense of where that bar is than I do. So if you're happy to roll this out, then I don't want to block you. I appreciate you've done some manual testing and let the branch soak in Integration for a bit.
> 
> (In case you're wondering what I would do if it were my change: I'd probably split up and simplify pollAndReload to make it a more obvious "if shouldReload() then reload" loop, rip out those dubious pre-existing/copied `recover`s and the anonymous functions that are only there because of them, and add some tests for all the stuff I'd touched - which would then be easier/possible to write once the nontrivial stuff is no longer buried in anon functions. Nothing stopping us doing that later, of course.)

Thanks for this Chris. I felt like we might have wanted to get this out of the door as it has been around for a little while now, but I hear what you say and I think you're right that it could benefit from some tests. I was initially happy to use the existing tests as they cover the behaviour of reloading routes and they caught quite a few issues during development, which gave me confidence that they had good coverage. Still, it probably is worth spending a bit more time adding some additional tests (there are currently no unit tests around `Router` for example, though as documented in the README, this was a design choice at the time `Router` was written). I'll get working on this and push up another commit once done - it might not cover everything you've mentioned but I think it should at least make a good start.

PR #203, opened 2021-04-27
Reduce Router log output

karlbaker02:
This PR reduces Router's log output, switching some of the `logInfo` calls that were added in https://github.com/alphagov/router/commit/678ef634657885e4eb552954ef89dcc29b47193c for `logDebug` calls instead. Alongside this, the Dockerfile has been updated to remove the `DEBUG` environment variable, which if set, will cause `logDebug` to write to output.

The result of these changes can be seen in the graph below, where they have been applied to integration:

Screenshot 2021-04-27 at 10 52 37

We are now only logging output when each Router instance starts up and when each Router instance is reloading.


fredericfran-gds (approved):
looks good

bilbof (approved):
Thanks @karlbaker02 👍

PR #204, opened 2021-04-27
Increase StartupDelay during deployments

bilbof:
#### Motivation

We see 503 errors when deploying router - this change hopes to reduce the number of errors we see.

The Router child process is returning a 503 error during deploys since the child process routing table is still empty when it receives its first request. We have recently started monitoring this error case: https://sentry.io/organizations/govuk/issues/2311515746/.

I think the routing table is taking longer to populate as more routes are added. 5 seconds may have been sufficient time to wait for the child process to get 'hot' in the past, but it now needs a bit longer.

#### Implementation

This overrides [the default startup delay](https://github.com/alext/tablecloth/blob/9a28c0f651e7faeb01e0381f263d1ad273cfc188/manager.go#L20) set in the Tablecloth library.

This is a small patch rather than fixing the underlying issue. The more robust fix will come when we move to hosting router in ECS, where we'll do blue-green deploys and remove the Tablecloth dependency.

I experimented deploying with a startup delay of both 30s and then 60s, and we see fewer errors when waiting for 60s. The chart below shows 503 errors served by staging cache machine's Nginx during deploys (deploy start times indicated by the green annotations):

Screenshot 2021-04-27 at 15 16 32

Please note that we won't see the benefit of this change until _after_ it is deployed. We'll still see the same error rate the next time we deploy.

### DIFF
--- a/main.go
+++ b/main.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"runtime"
 	"sync"
+	"time"
 
 	"github.com/alext/tablecloth"
 	"github.com/alphagov/router/handlers"
@@ -75,6 +76,7 @@ func logDebug(msg ...interface{}) {
 
 func catchListenAndServe(addr string, handler http.Handler, ident string, wg *sync.WaitGroup) {
 	defer wg.Done()
+	tablecloth.StartupDelay = 60 * time.Second
 	err := tablecloth.ListenAndServe(addr, handler, ident)
 	if err != nil {
 		log.Fatal(err)


karlbaker02:
I'm wondering if this will impact our app health checks for Router - if the start-up delay set in Tablecloth to start serving requests exceeds that of the health check used by AWS to monitor health of the `cache` machine and particularly that of the Router app, we _could_ end up in a scenario where we're never healthy as AWS tears the machine down, spins up a new machine and the process repeats.

bilbof:
Good thought. I don't think this will affect the health checks during deploys.

The parent process continues to serve requests until the child process is considered to have started (currently 5s, will be 60s). This means that the [liveness health check](https://github.com/alphagov/router/blob/811ed82d0ecc23b8077ed5dd2c460ccdfe2808ab/router_api.go#L36-L44) will be served by the existing process for a bit longer. This is why we see fewer errors with the `StartupDelay` increased - the parent process is serving requests (including health checks) while the child starts up.

I might expect to see the health checks failing if we _reduced_ the startup delay for the child process, perhaps to 0s. That could create the condition that no process would be available to serve the health check requests.

Also, this change will only affect router updates (not new machines), since `upgradeServer` only runs [when a `SIGHUP` is received](https://github.com/alext/tablecloth/blob/9a28c0f651e7faeb01e0381f263d1ad273cfc188/manager.go#L201-L206).

sengi:
That's a good question about the health checks. It won't matter for our existing setup on EC2 because the autoscaling group is only using the basic EC2 healthcheck; it's not configured to use the ELB healthcheck.

```
$ aws autoscaling describe-auto-scaling-groups --auto-scaling-group-names blue-cache
...
            "HealthCheckType": "EC2",
            "HealthCheckGracePeriod": 60,
```

PR #207, opened 2021-05-07
Remove disused checksum feature

sengi:
This essentially reverts #56.

AFAICT the checksum is not used anywhere (even in monitoring), but it recently started causing a real problem because it's asking MongoDB to sort the routes collection.  Apart from being inefficient, this runs into a scaling limitation of Mongo and led to route reloads failing under certain circumstances with errors like `Runner error during find: Overflow sort stage buffered data usage of 33554700 bytes exceeds internal limit of 33554432 bytes`.

We could alternatively add indices on those fields, but it's much better to remove the disused feature. This way we reduce the maintenance burden and maybe improve scalability a bit by eliminating unnecessary work on reloads.

Tested: Ran the testsuite locally:
```
$ make test_with_docker
...
Ran 103 of 106 Specs in 58.293 seconds
SUCCESS! -- 103 Passed | 0 Failed | 2 Pending | 1 Skipped
--- PASS: TestEverything (58.36s)
PASS
ok  	github.com/alphagov/router/integration_tests	58.710s
```

The `2 Pending | 1 Skipped` are the same with and without this change ("high throughput requires elevated ulimit", "connect timeout requires firewall block rule" and "should contain router internal server error metrics", respectively - all unrelated to this change).

The publishing-e2e-tests failure seems unrelated (`Container(s) finder-frontend were unhealthy after 60 seconds`).

[Trello card](https://trello.com/c/u0Grszrz/500)

### DIFF
--- a/integration_tests/performance_test.go
+++ b/integration_tests/performance_test.go
@@ -94,7 +94,7 @@ var _ = Describe("Performance", func() {
 		}
 	})
 
-	Describe("many concurrent (slow) connectons", func() {
+	Describe("many concurrent (slow) connections", func() {
 		if os.Getenv("RUN_ULIMIT_DEPENDENT_TESTS") != "" {
 			var (
 				backend1 *httptest.Server
--- a/integration_tests/reload_api_test.go
+++ b/integration_tests/reload_api_test.go
@@ -1,8 +1,6 @@
 package integration
 
 import (
-	"crypto/sha1"
-	"fmt"
 	"time"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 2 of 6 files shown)


thomasleese (approved):
Code changes look good, nice to remove an unused part of router, thanks!

sengi:
Rebased just to reword the main commit to correct an inaccuracy about the indices (they're not actually missing - at least not most of the time - but we suspect that when the database is being restored as part of the overnight sync in the staging/integration environments the index is effectively missing while it's being rebuilt and that's when the errors have been happening). Thanks @thomasleese for pointing that out.

bilbof (approved):
Great work!

PR #209, opened 2021-05-11
Rewrite README to be consistent with other repos

benthorner:
https://trello.com/c/Psf7EzPB/204-move-away-from-startupsh

Please see the commit messages for more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,19 +52,11 @@ This project uses [Go Modules](https://github.com/golang/go/wiki/Modules) to ven
 
     go mod vendor
 
-### Metrics
-
-Router exposes metrics in [Prometheus](https://prometheus.io) format using the
-[Go prometheus client
-library](https://godoc.org/github.com/dnesting/client_golang/prometheus).
-
-See commit b443d3dd9cf776143eed270d01bd98d2233caea6 as an example for how to
-add a metric.
-
 ### Further documentation
 
 - [Data structure](docs/data-structure.md)
 - [Original thinking behind the router](https://gdstechnology.blog.gov.uk/2013/12/05/building-a-new-router-for-gov-uk)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -18,22 +18,15 @@ Some of the thinking behind the router is documented in [this 2013 blog post][po
...
-a `$GOPATH` set up, you can simply run `go get github.com/alphagov/router` to
-pull the source into your workspace.
+Recommended reading: [How to Write Go Code](https://golang.org/doc/code.html)
 
-Once you have the code set up correctly, building and running the application
-is simple:
+You can use the [GOV.UK Docker environment](https://github.com/alphagov/govuk-docker) to run the application and its tests with all the necessary dependencies. Follow [the usage instructions](https://github.com/alphagov/govuk-docker#usage) to get started.
 
-    make
-    ./router -h
```
ChrisBAshton, line 53: Nice. Might be worth noting in the commit that we don't lose any information by deleting this section - [govuk-docker already runs `./router` via govuk-docker-up](https://github.com/alphagov/govuk-docker/blob/51e040b25d738f8e832b39624ece5fff697c3b96/projects/router/docker-compose.yml#L40). (I suspected that was the case but wanted to double check and make sure, given this is a unique app).

ChrisBAshton (approved):
LGTM 👍 nice job.

README.md
benthorner: Sure, I've done that, and also suggested we can reinstate them if non-Docker dev is unclear/desirable in future. However, as I discovered when doing this PR, it's important we push people to use GOV.UK Docker, as otherwise it can get broken unintentionally, which I've explained in this issue: https://github.com/alphagov/router/issues/210

Anyway, I'll merge this now, and you're welcome to deploy 😄.

PR #219, opened 2021-07-07
Fix cardinality of router_redirect_handler_redirect_total metric.

sengi:
This was OOMing Prometheus in prod and using taking up most of the long-term storage.

The fix is not to try to include the redirect URL as a label. Prometheus isn't designed for such high-cardinality metrics.

Tested: `Ran 103 of 106 Specs in 59.253 seconds` `SUCCESS! -- 103 Passed | 0 Failed | 2 Pending | 1 Skipped`

### DIFF
--- a/handlers/metrics.go
+++ b/handlers/metrics.go
@@ -13,7 +13,6 @@ var (
 		[]string{
 			"redirect_code",
 			"redirect_type",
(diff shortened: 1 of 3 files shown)


smford (approved):
Looks good to me.

PR #233, opened 2021-11-09
Remove ECR publish on PR

nsabri1:
After some discussion within the team, we have decided that its not necessary to have Github actions CI run on PR's for now.

What should happen on PR's is something we will visit again at a later date.


sengi (approved):
Good call - better safe than sorry! We can always put these hooks back once we've had a proper security review of all the CI stuff.

PR #253, opened 2022-02-18
Set the default header timeout to what we're actually using everywhere.

sengi:
We're overriding this from 15s to 20s in all environments, so let's just set the default to the value we want and save a bit of noise in configuration files.

See https://github.com/alphagov/govuk-puppet/blob/main/modules/govuk/manifests/apps/router.pp#L59


nsabri1 (approved):
LGTM

PR #284, opened 2022-08-12
Enable CI in GitHub Actions

theseanything:
This enables a GitHub Action workflow to run CI using a re-usableworkflow template in govuk-infrastructure repo. This is the initial set up to test the reusable workflow.

This PR also prevents the deploy workflow from being run before the application tests have been completed. This is to ensure that the commit passes application tests before being deployed.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,36 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - "Jenkinsfile"
+      - ".git**"
+  pull_request:
+
+jobs:
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,36 @@
...
+          go-version: '1.19.0'
+          cache: true
+
+      - name: Setup MongoDB
+        run: |
+          make start_mongo
+
+      - name: Build
+        run: |
+          make clean build
```
sengi, line 32: Is the `clean` target of any practical use here, given that it's running with a freshly checked-out tree?

sengi (approved):

.github/workflows/ci.yaml
theseanything: Good catch - removed it!

PR #347, opened 2023-03-27
Remove Jenkinsfile

Tetrino:
It's deprecated after the move to EKS/Kubernetes. The file still attempts to run a Jenkins job, which is no longer required and will eventually error every time as we remove the repos from that system.


Tetrino:
Closing after conversations with Sean Rankine over what needs doing here.

PR #349, opened 2023-03-31
Rename test to be more descriptive

Tetrino:
Previously it was simply named "test" which wasn't ideal for moving to EKS. This renaming allows us to properly reference it in the github actions setup over in govuk-saas-config.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,8 @@ on:
   pull_request:
 
 jobs:
-  test:
+  test-go:
+    name: Test Go
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -11,7 +11,8 @@ on:
   pull_request:
 
 jobs:
-  test:
+  test-golang:
+    name: Test GoLang
```
theseanything, line 15: You can probably simplify to "go", instead of golang (as by context most developers would understand its the language). 
```suggestion
  test-go:
    name: Test Go
```

theseanything (approved):

PR #353, opened 2023-04-17
Add ADR for porting router and router-api to PostgreSQL

chao-xian:
Rendered page


docs/ADR/001-porting-to-postgresql.md
```diff
@@ -0,0 +1,10 @@
+# ADR 001 - How we migrate Router/Router-API from Mongo to Postgres
+
+* Spin up a completely separate stack of router and router-api (govuk-docker first) that will talk to a PostgreSQL db only (as forks)
+* Provision in EKS a new PostgreSQL db
+* Provision the new p-router/p-router-api apps
+* Amend messages to router-api to be duplicated to p-router-api within the gds-api-adapters gem
```
leenagupte, line 6: This should be in content-store. The gds-api-adapters gem is just helper methods, it shouldn't have "business" logic in it.

leenagupte: So I'd split this into two steps. One to add helper methods to gds-api-adapter to call `p-router-api` (this still needs a better name!) and the next step to call the new helper methods from content store.

chao-xian: We did consider doing it in content-store, but felt it much easier to implement and manage (temporarily) in the gem.

Good point on the splitting of steps.

Tetrino: @chao-xian the suggestion from me to do the logic in the API (as small as that logic is) was based on the theory that it's one solid point of entry for everything. @leenagupte if content store is the only location that actually accesses these helper calls, we could do it there.

However to clarify, there is no business logic as such in the API adaptor in this plan, we simply append a call to another helper method inside the existing one (as is the plan).

leenagupte: Content Store is the only place that calls router / router-api. You'd be making a bit of a mess in gds-api-adapters if you try to have the router helpers call two places as you'd need to some work to get it to figure out what the endpoint is.

KludgeKML: My opinion is that either the API or the content store way of doing it would be fine, but the question of where/when to report errors probably makes the decision. Basically, if we do this splitting we need to be able to know about the following error cases:

-

chao-xian:
@aldavidson has suggested running a proxy to duplicate the requests and to be able to log - which is a good idea it seems

chao-xian:
@leenagupte yes he's suggesting a new app, and we can use that same proxy to duplicate our calls too

chao-xian:
Link to @aldavidson 's spike https://github.com/alphagov/content-store/pull/1062/

aldavidson:
Hi both - yes, we're looking at a simple proxy in front of two copies of content-store (one running Mongo, one running Postgres) that would forward all requests to both, and log any differences in the responses.

That will let us dual-run for a while, and build up a decent history of database backups before switching. Also lets us manage the switchover in phases (Mongo primary, PG secondary -> PG primary, Mongo secondary -> just PG -> no more proxy). It's based on the approach taken by [The Guardian when they ported their CMS from Mongo to PG](https://www.theguardian.com/info/2018/nov/30/bye-bye-mongo-hello-postgres). 

If you're considering the same, we'll hopefully be able to write the proxy in such a way that it's usable for both migrations

chao-xian:
See also https://github.com/alphagov/govuk-rfcs/pull/158

chao-xian:
We've started some spikes:

* https://github.com/alphagov/router-postgres
* https://github.com/alphagov/router-api-postgres

chao-xian:
See also from Al:

* https://github.com/alphagov/content-store-proxy
* https://github.com/alphagov/govuk-docker/pull/656

leenagupte (commented):
This sounds good. I think the only thing missing is any risks or consequences.

docs/ADR/001-porting-to-postgresql.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Context
+
+We need to migrate router (and [router-api](https://github.com/alphagov/router-api)) off of MongoDB as its datastore. The majority of GOV.UK platform is on PosgreSQL and there is no longer a strong enough case for being on MongoDB for router data.
+
+Even the [content-store is migrating from MongoDB to PostgreSQL](https://github.com/alphagov/govuk-rfcs/pull/158).
+
+This ADR is to outline our decisions and rationale on the steps we take for that migration:
+
+### The outline
```
leenagupte, line 11: I think this is the "Decision".

docs/ADR/001-porting-to-postgresql.md
```diff
@@ -0,0 +1,47 @@
...
+* We update router and router-api and deploy the changes to the existing version
+
+### Risks
+
+* Big bang
+* Running 2 databases under 1 app
+* Double code/logic to handle that
+* Harder to clean up
+
+## Decision
```
leenagupte, line 39: Not everyone has access to the Trello cards, so if there is any really useful information it should be copied into the ADR. Otherwise, I don't think these links fit here, I think they belong as links from the outline.

leenagupte:
@chao-xian Sorry, it occurred to me later that the Decision and the outline were mislabelled.

docs/ADR/001-porting-to-postgresql.md
leenagupte: This should be called "Decision". And be a Level 2 header.

chao-xian: Sorry misunderstood you before - done

leenagupte (approved):

PR #364, opened 2023-07-10
Router migration to postgres

chao-xian:
## What

Adding support for using PostgreSQL instead of MongoDB.

## Why

The version of Mongo we are on is deprecated and it's unnecessary to be on Mongo. 

https://trello.com/c/YJro3BGj/1790-spin-up-a-separate-stack-of-router-l

(This PR supersedes https://github.com/alphagov/router-postgres/pull/3)

## How

Instead of using a polling system, we now use PSQL's Notify channels to inform Router instances that they need to be updated with new routes.

This allows us to move over entirely to Postgres while reducing a lot of the complexity required to ensure we had an up to date status on the router-api database.

Currently we still poll the notify channel every two seconds.


chao-xian:
Instead of removing the MongoDB we'll need to set a feature flag to use PostgreSQL. This commit will need reworking f6b49e91decfd477ba425bc7f4a83b2a348c692d

hannako:
Superseded by https://github.com/alphagov/router/pull/389

PR #366, opened 2023-07-14
Fix `make test` + other Makefile improvements.

sengi:
- Start Mongo in a more reliable / less flaky way.
- Make the backend connection timeout test not rely on an external firewall, so we can actually run the test automatically like all the others.
- Fix the instructions in the README about how to run the tests.
- Remove the defunct Jenkinsfile.
- Remove the show_metrics manual testing cruft from the Makefile.
- Add separate targets to run just the unit tests or just the integration tests.
- Split the unit tests and integration tests into separate steps in the GitHub Actions job. Still takes the same time in the success case but it fails much faster when a unit test fails, which helps a bit.
    - We can easily run the two testsuites in parallel as separate jobs, but I'm leaving that for another PR because we'll need to update the set of required checks in GitHub at the same time.


sengi:
I found out what was up with the `docker: Error response from daemon: application/vnd.docker.distribution.manifest.v1+prettyjws not supported.` problem that I ran into on my other machine (spoiler alert: it's another Docker Desktop setting 😂)

Turns out it can't read that antique manifest format if "Use containerd for pulling and storing images" is enabled under Settings -> Features in development.

So I've changed it back to Mongo 2.4 so we're using the same version that's in ye olde crufty prod so at least at least we're not tempting fate with version drift problems 😅

sengi:
Also made it print a message mentioning those two settings when it fails to start the mongo container, so hopefully there'll be at least something to signpost other poor souls in the right direction when they run into this.

dj-maisy (approved):
🚀

sengi:
@chao-xian I'll go ahead and merge this as we chatted out-of-band. Thanks both for your help on this!

PR #369, opened 2023-07-22
Add lint checks and fix preexisting lints.

sengi:
Add [golangci-lint](https://golangci-lint.run/) as a pre-merge check, with GitHub review integration etc. and a make target for running it locally.

Fix preexisting lints; see commit messages.

Recommend reviewing commit-by-commit.

### DIFF
--- a/integration_tests/router_support.go
+++ b/integration_tests/router_support.go
@@ -9,7 +9,9 @@ import (
 	"syscall"
 	"time"
 
+	// revive:disable:dot-imports
 	. "github.com/onsi/gomega"
+	// revive:enable:dot-imports
 )
 
 func routerURL(path string, optionalPort ...int) string {
--- a/integration_tests/router_support.go
+++ b/integration_tests/router_support.go
@@ -81,8 +81,10 @@ func startRouter(port, apiPort int, optionalExtraEnv ...envMap) error {
 func stopRouter(port int) {
 	cmd := runningRouters[port]
 	if cmd != nil && cmd.Process != nil {
-		cmd.Process.Signal(syscall.SIGINT)
-		cmd.Process.Wait()
+		err := cmd.Process.Signal(syscall.SIGINT)
+		Expect(err).NotTo(HaveOccurred())
+		_, err = cmd.Process.Wait()
+		Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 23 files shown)


sengi:
cc @chao-xian

chao-xian:
wow this is a big PR...

KludgeKML:
FWIW, it looks pretty reasonable. Given it's adding linting, all the changes look like the exact sort of thing linting would bring up, so if it's passing tests I feel pretty confident that it's a decent set of changes.

sengi:
It's just a lot of mostly small commits that need to go in together.

I've kept roughly to fixing one type of issue per commit because it'll be much easier to review that way (especially since we have very few people familiar with Go).

sengi:
Btw my overall aim here is to get the Router codebase into better shape so that the Postgres project has a better chance of going smoothly. This should make it a lot easier and safer to work on.

chao-xian:
This mostly looks ok to me but I'm not a GoLang developer so I'd suggest getting another dev who has that level of familiarity to check as well

sengi:
Lol, thought I'd fixed the last of the flaky tests in #366 but apparently not 🙃 Let's try that again.

sengi:
Doesn't seem to be new flakiness (or if it is, it's quite a rare flake). So I think we're good.

chao-xian:
@sengi can another Platform Engineer such as @dj-maisy or @theseanything review and approve?

integration_tests/router_support.go
```diff
@@ -34,6 +34,7 @@ func reloadRoutes(optionalPort ...int) {
 	resp, err := http.Post(fmt.Sprintf("http://127.0.0.1:%d/reload", port), "", nil)
 	Expect(err).NotTo(HaveOccurred())
 	Expect(resp.StatusCode).To(Equal(202))
+	resp.Body.Close()
```
theseanything, line 47: good catch!

sengi: ah I can't take the credit, there's [a linter for this](https://github.com/golangci/golangci-lint/blob/master/pkg/golinters/bodyclose.go) :D

theseanything (approved):
Seems sensible to me!

PR #372, opened 2023-07-26
Eliminate patch version update toil.

sengi:
We always want to use the latest patch version of our pinned minor version of Go by default.

This eliminates the toil of keeping the patch version up-to-date and, more importantly, eliminates human forgetfulness from the security update path. We don't want critical security updates to require human intervention.

In the extremely unlikely event of a golang patch release causing a regression, it's still trivial to pin the patch version until it's fixed.

This also stops us pinning the Go version in 3 different places, which was getting rather silly. This gets us down to just `go.mod` and `Dockerfile`, both of which should now be covered by Dependabot.

Remove the long-winded procedure from the README and make Dependabot update `Dockerfile` on the same schedule as `go.mod` so they don't go out of sync. (Not that it really matters if they do, as long as Dockerfile is up-to-date. The [`go` property in go.mod](https://go.dev/doc/modules/gomod-ref#go) represents the *minimum* compatible Go version rather than the version that it "should" be build with or anything like that.)


sengi:
cc @chao-xian

chao-xian:
This seems to make sense to me

theseanything (approved):

PR #373, opened 2023-07-26
Remove tablecloth library as it's no longer needed.

sengi:
Now that we're finally containerised, we no longer need this library.

Which is lucky for us, because https://github.com/alext/tablecloth/ was archived in February and the last commit was 4 years ago.


sengi:
cc @chao-xian

sengi:
Ooh nice, the gosec linter doesn't like us using the default ServeMux. Hah. Could just make it a TODO but since it's come up now and it should be an easy fix, I'll try to get it in this PR.

theseanything (approved):
Nice one!

PR #376, opened 2023-07-28
Don't skip the high-concurrency tests.

sengi:
As of Go 1.19, the `os` package [sets `RLIMIT_NOFILE`](https://cs.opensource.google/go/go/+/go1.19.9:src/syscall/rlimit.go) to the maximum available on init, so we don't even need to call `os.Setrlimit()` ourselves to enable these tests.


sengi:
cc @chao-xian

theseanything (approved):

PR #377, opened 2023-07-28
More test improvements.

sengi:
- Delete some cruft from the metrics integration tests. All it was really doing was looking for a plausible 200 OK response for `/metrics`, so make it explicitly just do that and don't pretend to test the actual metrics. We should add proper unit tests (not integration tests) to check that counters are incremented when they're supposed to be (see [prometheus/testutil](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus/testutil)).
- Refactor the performance tests and make them generate a realistic request rate (50 qps for most tests and 500 qps for the high-throughput ones, instead of 1.7 qps and 50 qps respectively).

### DIFF
--- a/integration_tests/metrics_test.go
+++ b/integration_tests/metrics_test.go
@@ -5,16 +5,9 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("metrics API endpoint", func() {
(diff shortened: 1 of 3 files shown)


sengi:
cc @chao-xian

theseanything (approved):

PR #378, opened 2023-07-28
Upgrade to Ginkgo v2.

sengi:
Upgrade the test framework from Ginkgo v1 to Ginkgo v2.

Produces slightly nicer output and makes it possible to run tests in parallel in future.

The second commit is entirely generated by `go get -t -u ./... && go mod tidy && go mod vendor`.


sengi:
cc @chao-xian

theseanything (approved):

PR #380, opened 2023-07-29
Move Router code out of the main module etc.

sengi:
It's currently hard to unit test the methods of `Router` because it's tightly coupled with initialisation stuff in the main module. Moving it into its own module is the first step towards fixing this.

Make the integration tests more reliable and easier to maintain.

See commit messages for details.


Tetrino (approved):
Incredible work Chris 👍

theseanything (approved):
Nice clean up!!

PR #389, opened 2023-08-22
Router on postgres

Tetrino:
The underlying functionality is the same, but we are moving away from
mongo to postgres to ensure we can continue to maintain the app.

We no longer keep soft copies of the data read from the database,
instead reading each row from memory.

We no longer rely on comparing OpTime from Mongo changes, instead relying
on PSQL's LISTEN/NOTIFY process to notify us of changes. However, in order
to avoid overloading with constant update requests, we still poll
the update every two seconds.

https://trello.com/c/YJro3BGj


Makefile
```diff
@@ -27,14 +27,34 @@ lint:
...
+			--name router-postgres-test-db \
+			-e POSTGRES_HOST_AUTH_METHOD=trust \
+			-d \
+			-p 5432:5432 \
+			--user 'postgres' \
+			--health-cmd 'pg_isready' \
+			--health-start-period 5s \
+			postgres:14; \
+		echo Waiting for postgres to be up; \
+		until [ "`docker inspect -f '{{.State.Health.Status}}' router-postgres-test-db`" = "healthy" ]; do \
```
sengi, line 45: optional: if the docker or the container fails then this could get stuck for a long time (6h on GitHub Actions), so could so with `for _ in $(seq 30)`

sengi: sorry that was meant to be sent along with the review rather than as a one-off (I hit the wrong button)

lib/router.go
```diff
@@ -39,48 +38,38 @@ const (
...
 type Route struct {
-	IncomingPath string `bson:"incoming_path"`
-	RouteType    string `bson:"route_type"`
-	Handler      string `bson:"handler"`
-	BackendID    string `bson:"backend_id"`
-	RedirectTo   string `bson:"redirect_to"`
-	RedirectType string `bson:"redirect_type"`
-	SegmentsMode string `bson:"segments_mode"`
-	Disabled     bool   `bson:"disabled"`
+	IncomingPath sql.NullString
```
sengi, line 65: optional: it'd be neat if we could use plain old strings here and deal with the nullability in `loadRoutes` or wherever, then we could delete this struct's evil twin in `integration_tests/route_helpers.go`. (Would also mean you wouldn't need to deal with nulls in all the places that use struct Route.)

lib/router.go
```diff
@@ -249,83 +229,69 @@ func (rt *Router) reloadRoutes(db *mgo.Database, currentOptime bson.MongoTimesta
...
-	err = bson.Unmarshal(replicaSetStatusBytes, &replicaSet)
-	if err != nil {
-		return MongoReplicaSetMember{}, fmt.Errorf("router: couldn't unmarshal replica set status from MongoDB, skipping update (error: %w)", err)
-	}
-
-	currentInstance := make([]MongoReplicaSetMember, 0)
-	for _, instance := range replicaSet.Members {
-		if instance.Current {
-			currentInstance = append(currentInstance, instance)
+func (rt *Router) shouldReload(listener *pq.Listener) bool {
```
sengi, line 232: This is so much neater — I love it

localdb_init.sql
```diff
@@ -0,0 +1,67 @@
...
+  END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER backends_notify_trigger 
+AFTER INSERT OR UPDATE OR DELETE 
+ON routes 
+FOR EACH ROW
+EXECUTE PROCEDURE notify_listeners();
+
+CREATE TRIGGER rotues_notify_trigger
```
sengi, line 57: harmless typo
```suggestion
CREATE TRIGGER routes_notify_trigger 
```

localdb_init.sql
```diff
@@ -0,0 +1,67 @@
+DROP DATABASE IF EXISTS router_test;
```
sengi, line 1: It's kind of a shame we need this almost-but-not-quite-identical copy of the schema here, but I agree it's probably the right call in the interests of time rather than revamping the way the tests work to pull in the real schema from router-api 👍

sengi (approved):
Cool — I like the minimal approach.

sengi:
my optionals are genuinely optional btw — totally fine if you prefer not to do them (and no need to justify!)

localdb_init.sql
KludgeKML: Agree with @sengi 

The very slight downside of having router-api being the source of the database is that it makes everything in govuk-docker that depends on router now also dependent on router-api. But the advantages of having it as an ordinary rails migration outweighs that (plus the fact that router/router-api doesn't really work on govuk-docker at the moment anyway, and converting to PG might unblock that problem).

sengi: That just sounds like a govuk-docker bug/flaw tbh :) I hope next quarter we'll be able to make govuk-helm-charts portable enough to replace at least some govuk-docker use cases in local development / testing.

PR #424, opened 2024-01-20
Redirect all-caps article slugs to lowercase.

sengi:
Move this redirect functionality from nginx into Router, so that we can finally drop Perl from the nginx config.

We'll need to push this further down into the backend ("frontend") apps when we get rid of Router, but there's still value in eliminating Perl from the request path sooner rather than later.

Also some trivial style cleanup in proxy_function integration tests.


sihugh (approved):
🔽

sengi:
Thanks @sihugh!

PR #426, opened 2024-01-22
Add a configurable default backend.

sengi:
For all requests that don't match any path or path prefix in the database, forward the request to ROUTER_DEFAULT_BACKEND_URL. This defaults to `government-frontend`, which is where 93% of routes go.

This should allow shrinking the route database by an order of magnitude as well as making it a little easier to eliminate Router in the long run.

We no longer respond 503 to all requests when the route database is empty. This feature was presumably conceived as a kind of lameduck mode (albeit somewhat buggy). We're almost certainly better off without it now that an empty database doesn't mean we're completely broken.


theseanything (approved):
Thanks Chris! Agree, this seems like a sensible change. I'm pretty sure that Rails will return 404 for a route it doesn't have - but I think government-frontend has [wildcard route that catches everything](https://github.com/alphagov/government-frontend/blob/7855ff9d58d75b88c35085cbf33fb6aa177e9871/config/routes.rb#L26) and then gets handled by the content-item controller. I think there is assumption that there must be a content item in the Content Store if request is routed from router - so could potentially just raise an exception and return a 503 instead. https://github.com/alphagov/government-frontend/blob/7855ff9d58d75b88c35085cbf33fb6aa177e9871/app/controllers/content_items_controller.rb#L26

We might have to fix up some of the logic in government-frontend before rolling this out.

sengi:
Ohh good shout, thanks! I'll go explore that before rolling this out 🙇

sengi:
Did some exploratory testing in the staging environment and all seems happy apart from organisations pages like these:
- https://www.gov.uk/government/organisations/hm-revenue-customs/contact/excise-enquiries
- https://www.gov.uk/government/organisations/government-digital-service/about

I'll set this back to draft while I figure out what's going on there.

sengi:
Ughhh it's because those have routes that shadow shorter prefixes that go elsewhere 😫 (why GOV.UK whyyyyy)

```js
production:PRIMARY> db.routes.find({incoming_path: '/government/organisations/hm-revenue-customs'})
{
  "_id" : ObjectId("5ae1e0b545a5a01908a01215"),
  "backend_id" : "collections",
  "disabled" : false,
  "handler" : "backend",
  "incoming_path" : "/government/organisations/hm-revenue-customs",
  "route_type" : "prefix",
  "updated_at" : ISODate("2023-05-15T16:08:31.165Z")
}
production:PRIMARY> db.routes.find({incoming_path: '/government/organisations/hm-revenue-customs/contact/excise-enquiries'})
{
  "_id" : ObjectId("546dcf7e45a5a0a5a2000121"),
  "backend_id" : "government-frontend",
  "disabled" : false,
  "handler" : "backend",
  "incoming_path" : "/government/organisations/hm-revenue-customs/contact/excise-enquiries",
  "route_type" : "exact"
}
```

sengi:
Easy enough to just put a fancier filter on the `deleteMany()` one-off cleanup ofc, but it's disappointing if we're gonna be stuck with some annoying corner cases in government-frontend where it still has to poke at router-api 🫤

PR #440, opened 2024-03-05
Update toolchain to Go 1.22 and update dependencies.

sengi:
- Update the build toolchain from Go 1.21 to 1.22.
- `make update_deps` to update vendored libraries.


sengi:
Huh. I guess I'll leave the min version constraint in go.mod at 1.21 for now then, until https://www.github.com/github/codeql/issues/15647 is fixed 😅

samsimpson1 (approved):

PR #445, opened 2024-04-03
triemux: make shouldRedirToLowercase more efficient by regex compilation+reuse

### LINKED ISSUE #444: triemux: (*Mux).ServeHTTP could be sped up by hoisting out and compiling shouldRedirToLowercasePath's regular expression
Noticed while reading through this code that we've got this pattern https://github.com/alphagov/router/blob/919496ae8533b51c14b8767eae770450136b57ee/triemux/mux.go#L68-L70 in which every invocation of (*Mux).ServeHTTP goes through that code hence a hot path https://github.com/alphagov/router/blob/919496ae8533b51c14b8767eae770450136b57ee/triemux/mux.go#L49-L52

## Inefficiency
That code above in every single invocation calls regexp.MatchString which firstly has to freshly compile the *regexp.Regexp object then discard it after a mere match, at many calls such as for what this code is used (for the country's digital services), this inefficiency consumes a whole lot of CPU and RAM; and you can verify with this simple benchmark
```go
package triemux

import "testing"

var sink any = nil

var paths = []string{
	"/GOVERNMENT/GUIDANCE",
	"/GoVeRnMeNt/gUiDaNcE",
	"",
	"1234",
	"abc1234",
	"abc1234DE",
	"                                                    ",
	"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa  aaaaaaaaaa",
}

func BenchmarkShouldRedirectToLowercasePath(b *testing.B) {
	b.ReportAllocs()

	for i := 0; i < b.N; i++ {
		for _, path := range paths {
			sink = shouldRedirToLowercasePath(path)
		}
	}

	if sink == nil {
		b.Fatal("Benchmark did not run!")
	}

	sink = nil
}
```

### CPU profile
Screenshot 2024-04-03 at 6 22 34 AM

### Memory profile
Screenshot 2024-04-03 at 6 23 47 AM

## Remedy
That regular expression can be hoisted out and compiled once then reused each time with the (*regexp.Regexp).MatchString method per
```diff
diff --git a/triemux/mux.go b/triemux/mux.go
index 57e09d1..b410b51 100644
--- a/triemux/mux.go
+++ b/triemux/mux.go
@@ -59,6 +59,8 @@ func (mux *Mux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	handler.ServeHTTP(w, r)
 }
 
+var reShouldRedirect = regexp.MustCompile(`^\/[A-Z]+[A-Z\W\d]+$`)
+
 // shouldRedirToLowercasePath takes a URL path string (such as "/government/guidance")
 // and returns:
 //   - true, if path is in all caps; for example:
@@ -66,8 +68,7 @@ func (mux *Mux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 //   - false, otherwise; for example:
 //     "/GoVeRnMeNt/gUiDaNcE" -> false (should forward "/GoVeRnMeNt/gUiDaNcE" as-is)
 func shouldRedirToLowercasePath(path string) (match bool) {
-	match, _ = regexp.MatchString(`^\/[A-Z]+[A-Z\W\d]+$`, path)
-	return
+	return reShouldRedirect.MatchString(path)
 }
 
 // lookup finds a URL path in the Mux and returns the corresponding handler.
```

## Improvement results
The benchmark shows a drastic performance hit in all dimensions
```shell
$ benchstat before.txt after.txt 
name                             old time/op    new time/op    delta
ShouldRedirectToLowercasePath-8    29.8µs ± 1%     0.8µs ±24%   -97.15%  (p=0.000 n=8+10)

name                             old alloc/op   new alloc/op   delta
ShouldRedirectToLowercasePath-8    33.7kB ± 0%     0.0kB       -100.00%  (p=0.000 n=9+10)

name                             old allocs/op  new allocs/op  delta
ShouldRedirectToLowercasePath-8       360 ± 0%         0       -100.00%  (p=0.000 n=10+10)
```

and now all the time is spent properly matching the string and not in trying to compile and then discard the regular expression object.

### CPU profile
Screenshot 2024-04-03 at 6 29 51 AM

### Memory profile
When the benchmark runs, that path no longer consumes any RAM
Screenshot 2024-04-03 at 6 31 19 AM

odeke-em:
This change makes shouldRedirToLowercase more efficient by at start-up time compiling the regular expression that shouldRedirToLowercase uses instead of compiling and discarding it each time on a hit, with an exhibit below of the performance improvement:

```shell
$ benchstat before.txt after.txt
name                             old time/op    new time/op    delta
ShouldRedirectToLowercasePath-8    29.8µs ± 1%     0.8µs ±24%   -97.15%  (p=0.000 n=8+10)

name                             old alloc/op   new alloc/op   delta
ShouldRedirectToLowercasePath-8    33.7kB ± 0%     0.0kB       -100.00%  (p=0.000 n=9+10)

name                             old allocs/op  new allocs/op  delta
ShouldRedirectToLowercasePath-8       360 ± 0%         0       -100.00%  (p=0.000 n=10+10)
```

Fixes #444

### DIFF
--- a/triemux/mux.go
+++ b/triemux/mux.go
@@ -59,6 +59,8 @@ func (mux *Mux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	handler.ServeHTTP(w, r)
 }
 
+var reShouldRedirect = regexp.MustCompile(`^\/[A-Z]+[A-Z\W\d]+$`)
+
 // shouldRedirToLowercasePath takes a URL path string (such as "/government/guidance")
 // and returns:
 //   - true, if path is in all caps; for example:
@@ -66,8 +68,7 @@ func (mux *Mux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 //   - false, otherwise; for example:
 //     "/GoVeRnMeNt/gUiDaNcE" -> false (should forward "/GoVeRnMeNt/gUiDaNcE" as-is)
 func shouldRedirToLowercasePath(path string) (match bool) {
-	match, _ = regexp.MatchString(`^\/[A-Z]+[A-Z\W\d]+$`, path)
-	return
+	return reShouldRedirect.MatchString(path)
 }
 
 // lookup finds a URL path in the Mux and returns the corresponding handler.


sengi:
Thanks, @odeke-em. Good catch!

triemux/bench_test.go
theseanything, line 1: Thanks for benchmarking! We've got existing performance tests that should catch regressions, could we remove this benchmark to keep the codebase less cluttered?

odeke-em: Alright, sure, let me do so.

odeke-em: Done!

sengi (approved):

sengi:
Just to let you know this is in production now. Thanks for contributing!

CPU usage graph showing similar usage before and after the rollout, perhaps a slight reduction

odeke-em:
Great to see the impact of it running in production. Thank you for the reviews @sengi @theseanything!

PR #490, opened 2024-10-28
Add ability to load triemux from Content Store

theseanything:
This PR adds the ability to build and use a triemux from Content Store's postgres database instead of MongoDB. Moving away from MongoDB, which is running an outdated and unsupported version, resolves multiple challenges, including the inability to shift workloads to ARM and difficulties in maintaining a local development environment. By leveraging PostgreSQL, the Content Store can become the single source of truth for routes, simplifying the system architecture and removing the dependency on MongoDB.

This change also paves the way for retiring the Router API, allowing for a leaner, more efficient architecture. Instead of querying an API endpoint on the Content Store, this approach directly queries the PostgreSQL database. This is beneficial because the large number of routes involved would add significant latency if serialized and deserialized through an API layer. A risk to note is that schema changes in the Content Store could impact route loading in Router, though such changes are infrequent, and ongoing work may eventually make it possible to remove Router altogether.

Traffic control is managed through an environment variable, CSMUX_SAMPLE_RATE, which accepts values between 0.0 and 1.0. This variable sets the sampling rate for serving requests via the triemux built from the Content Store. A value of 0.0 disables triemux loading from the Content Store entirely, while higher values allow gradual adoption, enabling canarying of traffic to minimize rollout risks.

The Content Store PostgreSQL database URL is specified through the CONTENT_STORE_DATABASE_URL environment variable. Route re-loading is triggered automatically by PostgreSQL’s LISTEN/NOTIFY mechanism, which responds to inserts, updates, or deletions in the content_items or publish_intents tables. https://github.com/alphagov/content-store/pull/1340

This resolves: https://github.com/alphagov/router/issues/497


lib/routes.go
```diff
@@ -0,0 +1,71 @@
...
+	Details      *string
+}
+
+func (route *CsRoute) backend() *string {
+	if route.SchemaName != nil && *route.SchemaName == "gone" && !route.gone() {
+		if route.BackendID != nil {
+			return route.BackendID
+		} else {
+			defaultBackend := "government-frontend"
+			return &defaultBackend
```
nimalank7, line 29: `defaultBackend` is only in scope for `else` and not used elsewhere. This can just be `return "government-frontend"`

lib/routes_test.go
```diff
@@ -0,0 +1,147 @@
...
+				route.Details = &details
+				Expect(route.gone()).To(BeFalse())
+			})
+		})
+
+		Context("when schema is not 'gone'", func() {
+			It("should return false", func() {
+				Expect(route.gone()).To(BeFalse())
+			})
+		})
```
theseanything, line 93: Add test for json deserialisation failure

lib/load_routes.go
```diff
@@ -106,3 +109,43 @@ func loadRoutesFromCS(pool PgxIface, mux *triemux.Mux, backends map[string]http.
...
+			logger.NotifySentry(logger.ReportableError{Error: err})
+		}
+		timer.ObserveDuration()
+	}()
+
+	logInfo("router: reloading routes from content store")
+	newmux := triemux.NewMux()
+
+	err := loadRoutesFromCS(pool, newmux, rt.backends)
+	if err != nil {
```
nimalank7, line 202: This can be shifted into a single line.

lib/load_routes.go
```diff
@@ -110,6 +113,47 @@ func loadRoutesFromCS(pool PgxIface, mux *triemux.Mux, backends map[string]http.
...
+	}
+
+	listener.Handle("route_changes", pgxlisten.HandlerFunc(
+		func(ctx context.Context, notification *pgconn.Notification, conn *pgx.Conn) error {
+			rt.CsReloadChan <- true
+			return nil
+		}),
+	)
+
+	err := listener.Listen(ctx)
```
nimalank7, line 142: Single ine

lib/load_routes.go
```diff
@@ -110,6 +113,47 @@ func loadRoutesFromCS(pool PgxIface, mux *triemux.Mux, backends map[string]http.
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (rt *Router) waitForReload() {
+	for range rt.CsReloadChan {
+		func() {
```
nimalank7, line 166: I think this can be removed as it duplicates functionality already in `reloadCsRoutes`

cs_integration_tests/http_request_helpers.go
```diff
@@ -0,0 +1,83 @@
...
+	return resp
+}
+
+func doHTTP10Request(req *http.Request) *http.Response {
+	conn, err := net.Dial("tcp", req.URL.Host)
+	Expect(err).NotTo(HaveOccurred())
+	defer conn.Close()
+
+	if req.Method == "" {
+		req.Method = "GET"
```
theseanything, line 58: Use HTTP package const for GET

cs_integration_tests/integration_test.go
```diff
@@ -0,0 +1,59 @@
...
+var cleanupPostgresContainer func()
+
+func TestEverything(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Integration test suite")
+}
+
+var _ = BeforeSuite(func() {
+	runtime.GOMAXPROCS(runtime.NumCPU())
+	var err error
```
nimalank7, line 21: Move into Go `if` statement

cs_integration_tests/metrics_test.go
```diff
@@ -0,0 +1,22 @@
...
+	Context("response body", func() {
+		var responseBody string
+
+		BeforeEach(func() {
+			resp := doRequest(newRequest("GET", routerURL(apiPort, "/metrics")))
+			Expect(resp.StatusCode).To(Equal(200))
+			responseBody = readBody(resp)
+		})
+
+		It("should contain at least one metric", func() {
```
nimalank7, line 18: The `BeforeEach` is redundant as it's just a single test

lib/routes.go
theseanything: Sorry, I realised it needs to return a string pointer. You need to create the variable separately and then return the reference.

lib/routes_test.go
theseanything: Added

lib/routes.go
```diff
@@ -0,0 +1,71 @@
...
+		return true
+	}
+	return false
+}
+
+func (route *CsRoute) redirect() bool {
+	return route.SchemaName != nil && *route.SchemaName == "redirect"
+}
+
+func (route *CsRoute) segmentsMode() string {
```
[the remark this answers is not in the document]
theseanything: Done

cs_integration_tests/integration_test.go
theseanything: I can't because needs to be in scope for use elsewhere in the function

nimalank7 (approved):
LGTM!

PR #501, opened 2024-12-04
Remove ability to load routes from MongoDB

theseanything:
This can now be removed as we load routes from Content Store.

Tested in Integration


lib/logcompat.go
```diff
@@ -13,9 +13,3 @@ func logWarn(msg ...interface{}) {
 func logInfo(msg ...interface{}) {
 	log.Println(msg...)
 }
-
-func logDebug(msg ...interface{}) {
```
nimalank7, line 17: How comes this is being removed?

theseanything: The function isn't used anywhere - and the compiler/linter complains when you have un-used functions. Also we should move to use a proper logging library like zerolog.

PR #502, opened 2024-12-06
Improve periodic reload

theseanything:
Currently, it reloads every minute regardless if a reload has occurred within that minute. This is because we never set the last reload time. This also improves the responsiveness by checking for a periodic reload every 5 seconds, instead of a minute (which could potentially mean a reload could take up to 2 minutes).

Renamed csLastReloadTime to csLastAttemptReloadTime for clarity.

Tested in integration


MuriloDalRi (approved):
LGTM

(Although there's a failing test and I don't understand enough about what's going on there)

PR #582, opened 2025-09-01
[DO NOT MERGE] Change all gone routes to render with frontend

AshGDS:
Related to https://github.com/alphagov/frontend/pull/4917

Trello card: https://trello.com/c/2071GASB/755-move-gone-route-from-government-frontend-to-frontend


AgaDufrat (approved):
LGTM. Good call about testing this in integration first 🍀 

If we are certain that all "gone" routes are rendered with frontend, we should definitely simplify the logic here later on.

PR #676, opened 2026-04-23
Add router-probe-backend backend which is loaded whether from the database or file

jfharden:
Add in a fixed `/__probe__` prefix route which will always route to the router-probe-backend backend.

Load this route in both the load from content store, and load from file startup paths

### DIFF
--- a/lib/load_routes.go
+++ b/lib/load_routes.go
@@ -104,6 +104,28 @@ func loadRoutes(pool PgxIface, mux *triemux.Mux, backends map[string]http.Handle
 	if err := rows.Err(); err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


AP-Hunt (approved):
This is good as is, but I can't help but feel like the tests would be cleaner if they weren't making the HTTP requests, and instead able to query `mux` for a route with a certain prefix.

jfharden:
@AP-Hunt I agree, but I explored the mux object and it doesn't appear to have any externally available methods/attributes which allow you to query what handlers it actually has set and I definitely didn't feel like rewriting large parts of router, nor the tests :)

PR #678, opened 2026-04-28
Add tests for signal handling

jfharden:
Router isn't handling shutdown signals at all, so when a SIGINT or SIGTERM arrives all open connections are immediately severed and router exits non-zero.

I have a fix, but testing for it is a bit complex, so this PR adds in signal handling tests for SIGINT and SIGTERM but leaves them set as Pending (since they fail against main due to the lack of signal handling).

You can test these by checking this branch out and on line 14 of `integration_tests/signal_handling_test.go` remove the X from `XDescribe`, then run the integration tests: `make integration_tests`

You'll get the following results:

```
 $ make integration_tests
...SNIP...
go test -race -v ./integration_tests
=== RUN   TestEverything
Running Suite: Integration test suite - /Users/jonathan.harden/gitdevel/github.com/alphagov/router/integration_tests
====================================================================================================================
Random Seed: 1777374964

Will run 116 of 116 specs
...SNIP...
------------------------------
• [FAILED] [0.122 seconds]
Signal Handling When receiving a interrupt [It] Exits successfully
/Users/jonathan.harden/gitdevel/github.com/alphagov/router/integration_tests/signal_handling_test.go:31

  [FAILED] Expected
      <bool>: false
  to be true
  In [It] at: /Users/jonathan.harden/gitdevel/github.com/alphagov/router/integration_tests/signal_handling_test.go:35 @ 04/28/26 12:16:06.096
------------------------------
• [FAILED] [0.136 seconds]
Signal Handling When receiving a terminated [It] Exits successfully
/Users/jonathan.harden/gitdevel/github.com/alphagov/router/integration_tests/signal_handling_test.go:31

  [FAILED] Expected
      <bool>: false
  to be true
  In [It] at: /Users/jonathan.harden/gitdevel/github.com/alphagov/router/integration_tests/signal_handling_test.go:35 @ 04/28/26 12:16:06.221
------------------------------
• [FAILED] [0.189 seconds]
Signal Handling When receiving a interrupt [It] Gracefully allows connections to complete
/Users/jonathan.harden/gitdevel/github.com/alphagov/router/integration_tests/signal_handling_test.go:39

  [FAILED] Unexpected error:
      <*errors.joinError | 0xc0003e2090>:
      Get "http://127.0.0.1:3171/signals": read tcp 127.0.0.1:65527->127.0.0.1:3171: read: connection reset by peer
      {
          errs: [
              <*url.Error | 0xc0000d2210>{
                  Op: "Get",
[...]

PR #679, opened 2026-04-28
Add graceful shutdown handling for router for both SIGTERM and SIGINT

jfharden:
~REQUIRES https://github.com/alphagov/router/pull/678 to be merged first~

This gives router the ability to gracefully shut down when it receives SIGINT or SIGTERM.

Upon receipt of them it will call server shutdown with a Cancellation context which has a deadline set to the maximum of the read/write timeouts for onward requests, this should give all connections time to either complete, or timeout (as they naturally would) before.

It will wait for both the api server and the routing server to complete shutdown (which includes timeouts, timeouts do not stop graceful shutdown, but are logged) before terminating, and when it does terminate that will be with a successful shutdown.

Previously when router got a SIGINT or SIGTERM all connections were severed (causing connection reset by peer errors) and router exited non-zero (which also stops the go cover tool from writing it's coverage information (aside: This is how I discovered our lack of graceful shutdowns))

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,13 +1,17 @@
 package main
 
 import (
+	"context"
+	"errors"
 	"flag"
 	"fmt"
 	"log"
 	"net/http"
 	"os"
+	"os/signal"
 	"runtime"
 	"strconv"
+	"syscall"
 	"time"
 
 	"github.com/alphagov/router/handlers"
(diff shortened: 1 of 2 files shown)


jfharden:
This issue was discovered when doing https://github.com/alphagov/govuk-infrastructure/issues/3938 and will contribute towards it (since this fix _also_ fixes the last of test coverage generation during integration tests)

main.go
```diff
@@ -83,16 +87,54 @@ func mustParseDuration(s string) (d time.Duration) {
 	return
 }
 
-func listenAndServeOrFatal(addr string, handler http.Handler, rTimeout time.Duration, wTimeout time.Duration) {
+func maxDuration(a time.Duration, b time.Duration) time.Duration {
+	if a < b {
```
AP-Hunt: Isn't this backward? `time.Duration` is an alias for `int64`, so surely it should be `a < b? b : a`

As an side, you could use `math.Max` instead of an if.

jfharden: Yeah, you are absolutely right 😮‍💨 I've got this backwards.

I can't use math.Max since that operates on floats, but since 1.21 golang has built in min and max functions (which I learned looking up "golang math.Max for ints" 😄).

I'll ditch this function entirely and just call `max(a, b) where i call this`

Thanks :)

jfharden: @AP-Hunt fixed :)

AP-Hunt (approved):

PR #681, opened 2026-04-29
Add probe endpoints to test internal router behaviour

jfharden:
This PR adds internal router routes for the builtin gone and redirect behaviours. It adds:

* `/__probe__/gone` (returns 410 Gone)
* `/__probe__/router-redirect` (returns 301 Moved Permanently to `/__probe__/redirected`)

In order to maintain the behaviour that when no routes are loaded router immediately responds 503 Service Unavailable all of the probe routes will only be loaded if other routes have already been loaded.

The check that decides whether to respond 503 is done in the core of every request and I don't want to add any latency doing a complex check of all Tries to verify they all begin /__probe__. So I've purposefully made the behaviour not to load the probe routes if no other routes are loaded (and I've included unit and integration tests verifying this to be the case)


dj-maisy (approved):
LGTM. 👍

PR #688, opened 2026-05-05
Manage github actions with renovate so it can keep golangci-lint version up to date

jfharden:
Currently we are not version controlling golangci-lint, which means at any time (such as last week with golangci-lint version 2.12) our tests start failing on main.

The golangci-lint github action [suggests using renovate](https://github.com/golangci/golangci-lint-action#using-renovate-to-update-the-golangci-lint-version) which will keep the action, and the version of golangci-lint up to date.

So this PR moves the management of github actions updates from dependabot to renovate, and adds a version pin for the version of golangci-lint to the previous version 2.11.4. This means renovate (when configured to manage this repo via [this other PR](https://github.com/alphagov/govuk-helm-charts/pull/4266)) will open a PR to bump golangci-lint and we can use that PR to fix the lint errors.

It also means future updates after that will be managed without having main broken, and we will have the same cooldown as other repos.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,4 +1,9 @@
 {
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+      "config:recommended",
+      "helpers:pinGitHubActionDigests"
+  ],
   "enabledManagers": ["github-actions"],
   "minimumReleaseAge": "10 days"
 }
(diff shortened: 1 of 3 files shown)


renovate.json
samsimpson1, line 1: This should extend the recommended config:
```json
{
    "$schema": "https://docs.renovatebot.com/renovate-schema.json",
    "extends": [
        "config:recommended",
        "helpers:pinGitHubActionDigests"
    ],
  "enabledManagers": ["github-actions"],
  "minimumReleaseAge": "10 days"
}
```

jfharden: @samsimpson1 done

samsimpson1 (approved):

PR #727, opened 2026-08-05
Change gone response codes from 410 to 460

KludgeKML:
- We're currently suffering from an architectural double-bind. We don't want router to render pages, so when it handles a gone it just returns a bare 410 status code with "410 gone", which nginx then rewrites to point to a prerendered error page from an S3 bucket.
- However, this rewrite also nobbles any page an app returns, meaning that custom gone pages have to return a 200 (not ideal, particularly because it causes The National Archives to continue archiving the page publicly).
- Since it changes least and is expecting unusual behaviour, make the change here in router - if we want a 410 to get a prerendered page, instead return status code 460. This code is unused, and likely not to be in the foreseeable future. The nginx layer can then catch 460s and rewrite them to 410s with the prerendered content.
- This needs a change in helm-charts [(an example in integration is https://github.com/alphagov/govuk-helm-charts/pull/4389)](https://github.com/alphagov/govuk-helm-charts/pull/4392) (DONE)

- [x] DO NOT MERGE BEFORE https://github.com/alphagov/govuk-helm-charts/pull/4392 is deployed!

https://gov-uk.atlassian.net/browse/PNP-10105


README.md
```diff
@@ -53,9 +53,9 @@ Once a request is matched to a route, Router uses the `schemaName` property to d
 There are 3 handler types to handle a request:
 1. **backend**: Reverse proxies the request to a backend application server
 2. **redirect**: Returns an HTTP `301` redirect to a new location
-3. **gone**: Returns an HTTP `410` Gone response for deleted content
+3. **gone**: Returns an HTTP `460` Gone response for deleted content.
 
-Note: some `Gone` routes are also handled by the `backend` handler.
+Note: 460 is not the standard gone status code. This is because the nginx layer intercepts the code and rewrites it to 410 (and sends a nice static error page from an S3 bucket). We can't do the same for 410 because we need the frontend app to also be able to return 410 status codes for custom gone routes, and if they did that the nginx layer would intercept them as well, replacing the custom 410 page with the static one.
```
nimalank7: Can you keep the "some `Gone` routes are also handled by the `backend` handler."? This is needed as some `Gone` routes are handled by the `backend` handler.

nimalank7 (changes requested):

README.md
KludgeKML: Have updated to make it more explicit that this is the case (and that the backend handler is in practise always Frontend).

nimalank7 (approved):
LGTM!

